PR #2, opened 2018-10-04
Add a script to replace /Users with nfs mount

jcwilson:
Aladdin uses minikube for local development which is a VirtualBox-based VM. Internally it mounts your `/Users` (or `/cygdrive/c/Users` for Cygwin users) directory to the root of the minikube filesystem. We have discovered that the `vboxsf` mounts are not as performant as NFS mounts and furthermore are not as reliable when it comes to detecting file changes made from the host. To address this, we replace the `vboxsf` mounts within minikube with NFS mounts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## What is Aladdin?
 
-Inspired by the Genie of the Lamp, Aladdin is a command line tool built by Fivestars to simplify kubernetes operations for clusters management and application development. We have done this by combining several command line tools including kops, minikube, docker, kubectl, helm, awscli, and git. 
+Inspired by the Genie of the Lamp, Aladdin is a command line tool built by Fivestars to simplify kubernetes operations for clusters management and application development. We have done this by combining several command line tools including kops, minikube, docker, kubectl, helm, awscli, and git.
 
 Use aladdin to:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -35,24 +35,54 @@ To set up, just clone the Aladdin GitHub repository to get started:
...
     $ aladdin config set config_dir /path/to/aladdin/configuration
 
+### NFS mounts
+
+Aladdin uses minikube for local development which is a VirtualBox-based VM. Internally it mounts your `/Users` (or `/cygdrive/c/Users` for Cygwin users) directory to the root of the minikube filesystem. We have discovered that the `vboxsf` mounts are not as performant as NFS mounts and furthermore are not as reliable when it comes to detecting file changes made from the host. To address this, we replace the `vboxsf` mounts within minikube with NFS mounts. However, this requires a bit of manual setup on the host machine before it will take effect.
+
+#### NFS Host Setup
+
+1. Install an NFS service on your host
+    1. **macOS** You should already have it available.
```
jcwilson: Is this a reasonable expectation for all macOS users? Or am I making an assumption that only holds true at Fivestars where we provision machines with certain settings?

agianey: I think we're good here

PR #6, opened 2018-10-10
Various improvements

jcwilson:
* Add the `-i` shorthand for the `--init` command-line option 
* Add the `--image` command-line option to allow one to specify the aladdin image to use for
  aladdin commands
* Avoid the need for `chmod 6000` on our host's `.ssh/config` file by mounting the hosts `~/.ssh`
  directory at `/root/_ssh` and copying the directory to `/root/.ssh` when the container starts.
  This way the ownership is transferred to the `root` user.
* Exit with an error code if the deploy command fails the "master branch" check
* Use simplified `apk add` functionality to avoid cache bloat

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM alpine:latest
 
-RUN apk -Uuv \
-    add \
+RUN apk add --no-cache \
     bash \
     bash-completion \
     bats \
@@ -14,8 +13,7 @@ RUN apk -Uuv \
     openssh-client \
     python3 \
     ncurses \
-    vim \
-	&& rm /var/cache/apk/*
+    vim
 
 WORKDIR /root/aladdin
 COPY ./commands/python/requirements.txt ./commands/python/requirements.txt
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Aladdin uses minikube for local development which is a VirtualBox-based VM. Inte
     1. **Windows** You should install Cygwin and the `unfsd` package
         * Take care when configuring the `unfsd` package as it can inadvertently remove login capabilities for your active user account.
         * Be sure to add firewall exceptions for the installed services (/usr/sbin/unfsd & /usr/sbin/rpcbind).
-1. Add entries to your `/etc/hosts` file
+1. Add entries to your `/etc/exports` file
     1. **macOS**
     ```
     echo "/Users -alldirs -mapall="$(id -u)":"$(id -g)" $(minikube ip)" | sudo tee -a /etc/exports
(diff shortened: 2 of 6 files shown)


commands/python/command/deploy.py
```diff
@@ -56,7 +57,7 @@ def deploy(project, git_ref, namespace, dry_run=False, force=False, force_helm=F
             logging.error(f'You are deploying hash {git_ref} which does not match branch '
                           f'{cr.check_branch} on cluster {cr.cluster_name} for project '
                           f'{project}... exiting')
-            return
+            sys.exit(1)
```
jcwilson: I think the command dispatch module ought to interpret non-0 or non-None return values as failures and exit the overall command in such a way that the calling shell gets an error exit code

aladdin-container.sh
```diff
@@ -85,6 +85,11 @@ function environment_init() {
     echo "CLUSTER_CODE = $CLUSTER_CODE"
     echo "NAMESPACE = $NAMESPACE"
 
+    # Copy our .ssh directory over. This is a special case due to
+    # SSH permision and ownership checks on the ~/.ssh files.
+    rm -rf $HOME/.ssh
+    cp -r $HOME/_ssh $HOME/.ssh
```
jcwilson: During my testing, non-`IS_LOCAL` invocations don't get into that `if` statement below, so I had to put this out here.

README.md
```diff
@@ -64,7 +64,7 @@ Aladdin uses minikube for local development which is a VirtualBox-based VM. Inte
     1. **Windows** You should install Cygwin and the `unfsd` package
         * Take care when configuring the `unfsd` package as it can inadvertently remove login capabilities for your active user account.
         * Be sure to add firewall exceptions for the installed services (/usr/sbin/unfsd & /usr/sbin/rpcbind).
-1. Add entries to your `/etc/hosts` file
+1. Add entries to your `/etc/exports` file
```
jcwilson: oops

Dockerfile
```diff
@@ -1,7 +1,6 @@
 FROM alpine:latest
 
-RUN apk -Uuv \
-    add \
+RUN apk add --no-cache \
```
jcwilson: I couldn't find `-u` in the docs, and I didn't think the `-v` was useful. Would be glad to add either of those back.

aladdin.sh
```diff
@@ -266,6 +266,10 @@ while [[ $# -gt 0 ]]; do
             NAMESPACE="$2"
             shift # past argument
         ;;
+        -i|--image)
```
agianey: we'll need to move the `-i` here to `--init` as well

PR #12, opened 2018-10-26
Copy the ~/.ssh directory into minikube so it can be shared with the container as root

jcwilson:


aladdin.sh
```diff
@@ -225,6 +235,13 @@ function enter_docker_container() {
     fi
 
     local aladdin_image="${IMAGE:-"$(jq -r '.aladdin.repo' "$ALADDIN_CONFIG_FILE"):$(jq -r '.aladdin.tag' "$ALADDIN_CONFIG_FILE")"}"
+    local ssh_src
+
+    case "$OSTYPE" in
+        cygwin*) ssh_src="/.ssh" ;;
+        *)       ssh_src="$(pathnorm ~/.aws)" ;;
```
agianey: should be `"$(pathnorm ~/.ssh)"`?

PR #20, opened 2018-12-07
Fix escape sequence for confirm_production prompt

swagatpatel:


aladdin.sh
```diff
@@ -174,7 +174,7 @@ function pathnorm(){
 function confirm_production() {
     if $IS_TERMINAL ; then  # if this is an interactive shell and not jenkins or piped input, then verify
         echo "Script is running in a terminal. Let us make user aware that this is production";
-        echo -ne '\e[31mYou are on production. Please type "production" to continue: \033[0m'; read -r
+        echo -ne '\033[;31mYou are on production. Please type "production" to continue: \033[0m'; read -r
```
swagatpatel: I noticed that when we went from internal to opensource, this output, which used to be red isn't so anymore.

it sounds like using `\033` is the more cross-platform friendly way to do this, according to this SO [link](https://stackoverflow.com/questions/5947742/how-to-change-the-output-color-of-echo-in-linux).

PR #24, opened 2019-02-12
[ALDN-94] Add pagination support

jarojasm95:
This handles `Exception('Does not handle truncated hostedzone')`

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53.html#Route53.Paginator.ListResourceRecordSets


abandali:
Looking at the error that originates from this [LOC](https://github.com/fivestars-os/aladdin/blob/master/commands/python/libs/aws/dns_mapping.py#L91-L92) it seems that this pagination approach should fix the issue

PR #26, opened 2019-10-09
Support new commands package pattern

jarojasm95:


agianey:
Is this still relevant since https://github.com/fivestars-os/aladdin/pull/44? or can we close this

PR #30, opened 2019-12-05
Add support for aws iam authenticator

agianey:
- [x] Add to README for how to use this

- [x] Add more comments to code

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:3.6
 
 RUN apk add --no-cache \
     bash \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:alpine
+FROM golang:1.13.5-alpine3.10
 
 RUN apk add --no-cache \
     bash \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.6
+FROM golang:alpine
 
 RUN apk add --no-cache \
     bash \
@@ -32,6 +32,8 @@ RUN	curl -L -o /bin/kops https://github.com/kubernetes/kops/releases/download/$K
 
 RUN	curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && cp linux-amd64/helm \
 	/bin/helm && chmod 755 /bin/helm && helm init --client-only
+
+RUN go get -u -v sigs.k8s.io/aws-iam-authenticator/cmd/aws-iam-authenticator
 
 ENV PATH="/root/.local/bin:/root:${PATH}"
 COPY . /root/aladdin
--- a/aladdin-container.sh
+++ b/aladdin-container.sh
@@ -118,6 +118,36 @@ function environment_init() {
 
 }
 
+function add_and_set_authentication_users() {
+    # If authentication is enabled, add the appropriate users to our kubeconfig
+    if $AUTHENTICATION_ENABLED; then
+        jq -r '.|keys[]' <<< "$AUTHENTICATION_ROLES" | while read name ; do
+            role_arn="$(jq -r --arg name "$name" '.[$name]' <<< $AUTHENTICATION_ROLES)"
+            _add_authentication_user_to_kubeconfig "$name" "$role_arn"
+        done
+        kubectl config set-context "$NAMESPACE.$CLUSTER_NAME" --cluster "$CLUSTER_NAME" --namespace="$NAMESPACE" --user "$AUTHENTICATION_ALADDIN_ROLE"
+    fi
+}
+
+function _add_authentication_user_to_kubeconfig() {
+    name="$1"
+    role_arn="$2"
+cat <<EOT >> $HOME/.kube/config
+- name: $name
+  user:
+    exec:
+      apiVersion: client.authentication.k8s.io/v1alpha1
+      args:
+      - token
+      - -i
+      - $CLUSTER_NAME
+      - -r
+      - $role_arn
+      command: aws-iam-authenticator
+EOT
+}
+
 source_cluster_env
 environment_init
+add_and_set_authentication_users
 exec_command_or_plugin "$@"
--- a/aladdin-container.sh
+++ b/aladdin-container.sh
@@ -106,9 +106,11 @@ function environment_init() {
         kubectl config set-context "$NAMESPACE.$CLUSTER_NAME" --cluster "$CLUSTER_NAME" --namespace="$NAMESPACE" --user "$CLUSTER_NAME"
         kubectl config use-context "$NAMESPACE.$CLUSTER_NAME"
 
+        add_and_set_authentication_users
+
         if $INIT; then
             kubectl create namespace --cluster $CLUSTER_NAME $NAMESPACE || true
-            helm init --upgrade --wait || true
+            _initialize_helm
             _replace_aws_secret || true
             $PY_MAIN namespace-init --force
         fi
@@ -116,6 +118,16 @@ function environment_init() {
 
     echo "END ENVIRONMENT CONFIGURATION==============================================="
 
+}
+
+function _initialize_helm() {
+    if $RBAC_ENABLED; then
+        kubectl -n kube-system create serviceaccount tiller || true
+        kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount=kube-system:tiller || true
+        helm init --service-account=tiller --upgrade --wait || true
+    else
+        helm init --upgrade --wait || true
+    fi
 }
 
 function add_and_set_authentication_users() {
@@ -149,5 +161,4 @@ EOT
 
 source_cluster_env
 environment_init
-add_and_set_authentication_users
 exec_command_or_plugin "$@"
--- a/aladdin-container.sh
+++ b/aladdin-container.sh
@@ -132,7 +132,7 @@ function add_and_set_authentication_users() {
 function _add_authentication_user_to_kubeconfig() {
     name="$1"
     role_arn="$2"
-cat <<EOT >> $HOME/.kube/config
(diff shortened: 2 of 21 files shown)


aladdin-container.sh
```diff
@@ -118,6 +118,35 @@ function environment_init() {
...
 }
 
+function add_and_set_authentication_users() {
+    if $AUTHENTICATION_ENABLED; then
+        jq -r '.|keys[]' <<< "$AUTHENTICATION_ROLES" | while read name ; do
+            role_arn="$(jq -r --arg name "$name" '.[$name]' <<< $AUTHENTICATION_ROLES)"
+            _add_authentication_user_to_kubeconfig "$name" "$role_arn"
+        done
+        kubectl config set-context "$NAMESPACE.$CLUSTER_NAME" --cluster "$CLUSTER_NAME" --namespace="$NAMESPACE" --user "$AUTHENTICATION_ALADDIN_ROLE"
+    fi
```
agianey: This function adds all the different roles to our kubeconfig

aladdin-container.sh
```diff
@@ -118,6 +118,35 @@ function environment_init() {
...
+    exec:
+      apiVersion: client.authentication.k8s.io/v1alpha1
+      args:
+      - token
+      - -i
+      - $CLUSTER_NAME
+      - -r
+      - $role_arn
+      command: aws-iam-authenticator
+EOT
```
agianey: Couldn't figure out how to leave this indented without breaking syntax

aladdin.sh
```diff
@@ -125,7 +125,14 @@ function enter_minikube_env() {
...
     value="$1"
-    jq -r ".$value" "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json"
+    output=$(jq -r ".$value" "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json")
+    if [[ "$output" == null ]]; then
+        output=$(jq -r ".$value" "$ALADDIN_CONFIG_DIR/default/config.json")
+    fi
+    if [[ "$output" == null ]]; then
+        output=$(jq -r ".$value" "$ALADDIN_CONFIG_DIR/config.json")
+    fi
+    echo "$output"
```
agianey: Change this function to check configurations in cluster, default, and overall config files

commands/python/requirements.txt
```diff
@@ -3,4 +3,4 @@ boto3==1.7.33
 cached-property==1.3.0
 colorlog==2.10.0
 jmespath==0.9.2
-kubernetes==5.0.0
+kubernetes==10.0.1
```
agianey: Needed to upgrade this because alpine is past python3.6, and the old kubernetes lib assigns to the async keyword

aladdin-container.sh
agianey: This kubeconfig file could alternatively just be in s3 and we could pull it down, rather than generated on the fly like it is here

Dockerfile
```diff
@@ -33,5 +33,7 @@ RUN	curl -L -o /bin/kops https://github.com/kubernetes/kops/releases/download/$K
 RUN	curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && cp linux-amd64/helm \
 	/bin/helm && chmod 755 /bin/helm && helm init --client-only
 
+RUN go get -u -v sigs.k8s.io/aws-iam-authenticator/cmd/aws-iam-authenticator
```
swagatpatel: does this need a version?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM golang:alpine
```
swagatpatel: should this also have a version?

commands/bash/container/change-permissions/change-permissions
```diff
@@ -0,0 +1,35 @@
...
+		positional arguments:
+		  permission            the permission you want to elevate to
+          
+		optional arguments:
+		  -h, --help            show this help message and exit
+	EOF
+}
+
+if [[ $# -eq 0 || "$1" == "-h" || "$1" == "--help" ]]; then
+    help
```
swagatpatel: the syntax highlighting points to `help` being a builtin bash command, maybe we name this `_help`?

aladdin-container.sh
swagatpatel: I think generation inline is fine.

commands/python/requirements.txt
swagatpatel: will this have to be coordinated with services that use aladdin + kubernetes? (like `loyalty`)

aladdin-container.sh
ptramsey: There's a way to do it but it involves using tabs for indentation which is likely to create issues in the future as I know I've run into code that works that way and not known I was breaking it by using spaces :\  ^EOT seems like a fine compromise to me.

ptramsey: You *could* indent just the initial cat command ...

aladdin.sh
```diff
@@ -123,9 +123,17 @@ function enter_minikube_env() {
 }
 
 function _extract_cluster_config_value() {
+    # Try extracting config from cluster config.json, default config.json, then aladdin config.json
     local value
     value="$1"
-    jq -r ".$value" "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json"
+    output=$(jq -r ".$value" "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json")
+    if [[ "$output" == null ]]; then
```
ptramsey: This seemed verbose, so I spent some time digging around in the jq man page because I was curious if there was another way to say this / to remove the actual 'null' value which feels awkward in a shell script.

So, it looks like we could do, e.g.,
```
# Find the result from the first input file that matches .$value
jq -enr "first(inputs | .$value // empty)" "$config1" "$config2" "$config3"
```
And that would echo the first match or nothing if nothing matched, and also return nonzero if there were no matches (so we could distinguish that case from empty string).

That said, now that I look at it, that might be more confusing for anyone who finds it in the future, and it's not more correct.

Well, now I know more about jq :man_shrugging:

commands/bash/container/change-permissions/change-permissions
```diff
@@ -0,0 +1,35 @@
...
+	# $AUTHENTICATION_ALLOWED_CHANGE_ROLES
+	if ! "$AUTHENTICATION_ENABLED"; then
+		echo >&2 "aladdin: error: authentication is not enabled currently. update your aladdin-config to enable"
+		exit 1
+	fi
+	if [[ "$(jq -r --arg role $1 '. | index($role)' <<< $AUTHENTICATION_ALLOWED_CHANGE_ROLES)" == null ]]; then
+		echo >&2 "aladdin: error: this permission does not exist, or you are not allowed to change to it"
+		exit 1
+	fi
+    kubectl config set-context "$NAMESPACE.$CLUSTER_NAME" --cluster "$CLUSTER_NAME" \
```
ptramsey: indentation doesn't match

commands/bash/container/change-permissions/change-permissions
ptramsey: or `usage`

commands/python/requirements.txt
ptramsey: alternatively, can we pin to older alpine?

aladdin.sh
ptramsey: Actually I take that back.  As it is right now, with the `-r` option, the string "null" is indistinguishable from an actual null.  The code I specified above is more correct.

ptramsey: EDIT: Updated the initial command to add quotes around the selector, to allow for keys with spaces in them

agianey: awesome.. this is way better

aladdin.sh
```diff
@@ -123,26 +123,38 @@ function enter_minikube_env() {
 }
 
 function _extract_cluster_config_value() {
+    # Try extracting config from cluster config.json, default config.json, then aladdin config.json
     local value
     value="$1"
-    jq -r ".$value" "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json"
+    jq -enr "first(inputs | .\"$value\" // empty)" "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json" \
```
ptramsey: you can remove the `-e` then you don't need the || true :)  the -e just makes it error on null but if we don't care about that case we can just have the default behavior of empty output on null

commands/python/requirements.txt
agianey: older alpine didn't have the right go version to install aws-iam-authenticator

Dockerfile
agianey: I left this without version since it kept throwing errors when i tried to pin it to the 0.4.0 (the latest)

commands/python/libs/k8s/ingress.py
```diff
@@ -26,26 +26,26 @@ def _create_ingress_service_tuples(services, dual_dns_prefix_annotation_name):
 
 
 def build_ingress(services, dns_suffix, dual_dns_prefix_annotation_name, ingress_info):
-    ingress = client.V1beta1Ingress()
+    ingress = client.ExtensionsV1beta1Ingress()
```
agianey: These changes are part of the kubernetes lib upgrade

aladdin.sh
```diff
@@ -123,26 +123,38 @@ function enter_minikube_env() {
 }
 
 function _extract_cluster_config_value() {
+    # Try extracting config from cluster config.json, default config.json, then aladdin config.json
     local value
     value="$1"
-    jq -r ".$value" "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json"
+    jq -nr --arg value "$value" "first(inputs | .$value // empty)" "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json" \
```
ptramsey: Discussed in PM but for posterity: needs single quotes around `first(....)` and square brackets around `$value`

PR #32, opened 2019-12-13
v1.15.6

jarojasm95:


commands/python/requirements.txt
```diff
@@ -1,6 +1,6 @@
-awscli==1.15.42
-boto3==1.7.33
+awscli==1.16.294
+boto3==1.10.30
 cached-property==1.3.0
 colorlog==2.10.0
 jmespath==0.9.2
-kubernetes==5.0.0
+kubernetes==11.0.0b2
```
agianey: Prefer to avoid beta versions (assuming that's what the b means), but if you've seen no issues, then let's leave it

jarojasm95: yeah same here, unfortunately this is the only version that supports our cluster version https://github.com/kubernetes-client/python#compatibility-matrix

PR #33, opened 2019-12-17
Fix nfs mount

jarojasm95:


scripts/install_nfs_mounts.sh
```diff
@@ -15,14 +15,14 @@ case "$OSTYPE" in
         ;;
 esac
 
-minikube_ip=$(minikube ip)
+host_ip="192.168.99.1"
```
jarojasm95: this is running inside minikube, so we need the host ip as seen from minikube

scripts/install_nfs_mounts.sh
```diff
@@ -15,14 +15,14 @@ case "$OSTYPE" in
...
 read -r -d '' bootlocal <<EOF ||:
 #!/bin/sh
 
 sudo systemctl start nfs-client.target
 sudo mkdir -p ${mount_dir}
-sudo busybox umount ${mount_dir} 2>/dev/null
-sudo busybox mount -t nfs -o tcp,rw,hard,noacl,async,nolock ${minikube_ip}:${export_dir} ${mount_dir}
+sudo umount ${mount_dir} 2>/dev/null
+sudo mount -t nfs -o tcp,rw,hard,noacl,async,nolock ${host_ip}:${export_dir} ${mount_dir}
```
jarojasm95: busybox was not working

jarojasm95:
Initially missed this because Virtualbox kept the mounts working

PR #34, opened 2020-01-14
INFRA-194: Allow rbac support

agianey:


aladdin-container.sh
```diff
@@ -118,6 +118,16 @@ function environment_init() {
 
 }
 
+function _helm_init() {
+    if $RBAC_ENABLED; then
+        kubectl -n kube-system create serviceaccount tiller || true
+        kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount=kube-system:tiller || true
+        helm init --service-account=tiller
+    else
+        helm init
```
ptramsey: why is the fallback not the same as the original code?  IE, is there a reason this isn't `helm init --upgrade --wait || true` ?

agianey: haha whoops. good catch

aladdin-container.sh
```diff
@@ -106,6 +106,8 @@ function environment_init() {
         kubectl config set-context "$NAMESPACE.$CLUSTER_NAME" --cluster "$CLUSTER_NAME" --namespace="$NAMESPACE" --user "$CLUSTER_NAME"
         kubectl config use-context "$NAMESPACE.$CLUSTER_NAME"
 
+        add_and_set_authentication_users
+
```
agianey: This is so that, in the case where the cluster isn't created, we don't try and do this

PR #35, opened 2020-01-23
ALDN-127: for get-dashboard-url use kops instead of kubectl

abandali:


commands/bash/container/get-dashboard-url/get-dashboard-url
```diff
@@ -2,12 +2,12 @@
 set -eu -o pipefail
 
 function get_dashboard_url {
-    base=$(kubectl cluster-info | grep "Kubernetes master" | cut -d '/' -f3)
-    if [[ $CLUSTER_CODE = "minikube" ]]; then
```
abandali: moved `base` inside the if and fixed the "=" in if condition

commands/python/main.py
```diff
@@ -83,7 +83,7 @@
                         help='Force initialization logic')
     parser.add_argument('--dev', action='store_true',
                         help='Mount host\'s aladdin directory onto aladdin container')
-    parser.add_argument('--image', help='Use the specified aladdin image (if building it yourself')
+    parser.add_argument('--image', help='Use the specified aladdin image (if building it yourself)')
```
abandali: nit grammatical fix

commands/bash/container/get-dashboard-url/get-dashboard-url
```diff
@@ -2,12 +2,12 @@
...
+    if [[ $CLUSTER_CODE == "minikube" ]]; then
+        base=$(kubectl cluster-info | grep "Kubernetes master" | cut -d '/' -f3)
```
agianey: wait, does this work? Should we change this output `please run "minikube dashboard" from your host` or something?

abandali: well, interesting point ... it works, in that it gives some output:
```
$ aladdin --dev get-dashboard-url
Including environment variables from script /root/aladdin-config/minikube/env.sh
START ENVIRONMENT CONFIGURATION=============================================
CLUSTER_CODE = minikube
NAMESPACE = default
Context "default.minikube" created.
Switched to context "default.minikube".
END ENVIRONMENT CONFIGURATION===============================================
192.168.99.100:30000
```

But that URL is bogus ... so yeah lemme update it with a message

PR #36, opened 2020-02-19
[ALDN-130] Allow running minikube with vm-driver=none

ptramsey:
This PR updates aladdin to allow running natively on the host's docker
 if available, by setting the new minikube.vm_driver preference in
 ~/.aladdin/config/config.json to 'none' (the default is 'virtualbox').
 This also adds a 'minikube.memory' preference to allow configuring
 the minikube VM's memory, if applicable.
 
 [ALDN-130](https://fivestars.atlassian.net/browse/ALDN-130)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN pip3 install --no-cache-dir -r ./commands/python/requirements.txt
 ARG KUBE_VERSION=1.15.6
 ARG KOPS_VERSION=1.15.0
 ARG HELM_VERSION=2.16.1
+ARG DOCKER_VERSION=18.09.7
 
 RUN	curl -L -o /bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v$KUBE_VERSION/bin/linux/amd64/kubectl \
 	&& chmod 755 /bin/kubectl
@@ -32,6 +33,9 @@ RUN	curl -L -o /bin/kops https://github.com/kubernetes/kops/releases/download/$K
 
 RUN	curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && cp linux-amd64/helm \
 	/bin/helm && chmod 755 /bin/helm && helm init --client-only
+
+RUN curl -L -o- https://download.docker.com/linux/static/stable/x86_64/docker-$DOCKER_VERSION.tgz | tar -zxvf - && cp docker/docker \
+    /usr/bin/docker && chmod 755 /usr/bin/docker
 
 RUN go get -u -v sigs.k8s.io/aws-iam-authenticator/cmd/aws-iam-authenticator
 
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -128,7 +128,7 @@ function check_or_start_minikube() {
         echo "Starting minikube... (this will take a moment)"
         set_minikube_config
 
-        _start_minikube
+        _start_minikube >/dev/null
 
         # Determine if we've installed our bootlocal.sh script to replace the vboxsf mounts with nfs mounts
         if ! "$(minikube ssh -- "test -x /var/lib/boot2docker/bootlocal.sh && echo -n true || echo -n false")"; then
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -116,8 +116,10 @@ function check_or_start_minikube() {
         # Determine if we've installed our bootlocal.sh script to replace the vboxsf mounts with nfs mounts
         if ! "$(minikube ssh -- "test -x /var/lib/boot2docker/bootlocal.sh && echo -n true || echo -n false")"; then
             echo "Installing NFS mounts from host..."
-            "$SCRIPT_DIR"/install_nfs_mounts.sh
-            echo "NFS mounts NOT installed"
+            if test $(minikube config get vm-driver) != "none"; then
+                "$SCRIPT_DIR"/install_nfs_mounts.sh
+                echo "NFS mounts installed"
+            fi
         fi
         echo "Minikube started"
     else
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -89,12 +89,18 @@ function check_and_handle_init() {
 }
 
 function set_minikube_config(){
-    local vm_driver=$(jq --arg DEFAULT "$DEFAULT_MINIKUBE_VM_DRIVER" -r '.minikube.vm_driver//$DEFAULT' "$ALADDIN_CONFIG_FILE")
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -33,6 +34,9 @@ RUN	curl -L -o /bin/kops https://github.com/kubernetes/kops/releases/download/$K
 RUN	curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && cp linux-amd64/helm \
 	/bin/helm && chmod 755 /bin/helm && helm init --client-only
 
+RUN curl -L -o- https://download.docker.com/linux/static/stable/x86_64/docker-$DOCKER_VERSION.tgz | tar -zxvf - && cp docker/docker \
```
ptramsey: We no longer depend on mounting the docker executable from the underlying system, as there's no guarantee that it will be statically linked (and therefore run inside our alpine/musl-based aladdin container).  Since we already directly download the docker executable in infra_k8s_check.sh, doing it here too seemed reasonable.

aladdin.sh
```diff
@@ -85,22 +89,54 @@ function check_and_handle_init() {
 }
 
 function set_minikube_config(){
+    local vm_driver=$(jq --arg DEFAULT "$DEFAULT_MINIKUBE_VM_DRIVER" -r '.minikube.vm_driver//$DEFAULT' "$ALADDIN_CONFIG_FILE")
+    local memory=$(jq --arg DEFAULT "$DEFAULT_MINIKUBE_MEMORY" -r '.minikube.memory//$DEFAULT' "$ALADDIN_CONFIG_FILE")
```
agianey: Is the expectation here that if someone wants to use a different vm driver/memory, they need to update their aladdin config? Currently I think the pattern is for users to not really touch their aladdin config, but I'm open to that pattern being changed.

aladdin.sh
```diff
@@ -85,22 +89,54 @@ function check_and_handle_init() {
...
+            minikube_cmd="sudo -E CHANGE_MINIKUBE_NONE_USER=true '$(which minikube)' start"
+
+        else
+            # On linux, /home gets mounted on /hosthome in the minikube vm, so as not to
+            # override /home/docker.  We still want the user's home directory to be in the
+            # same path both in and outside the minikube vm though, so symlink it into place.
+            minikube_cmd="$minikube_cmd &&\
+                          minikube ssh \"sudo mkdir '$HOME' && \
+                                         sudo mount --bind '${HOME/\/home//hosthome}' '$HOME'\""
+        fi
```
agianey: This part looks fine to me, but maybe beneficial to get @bbil to look at this section?

aladdin.sh
agianey: Maybe add support for cpu and disk size while you're in here? i think that will take care of the "most popularly" toggled minikube settings

ptramsey: Done.

ptramsey: Switched to using aladdin config set/get, rather than looking at a specific file.

aladdin.sh
```diff
@@ -86,21 +90,59 @@ function check_and_handle_init() {
 
 function set_minikube_config(){
     minikube config set kubernetes-version v$KUBERNETES_VERSION &> /dev/null
-    minikube config set vm-driver virtualbox &> /dev/null
-    minikube config set memory 4096 &> /dev/null
+
+    for key in vm_driver memory disk_size cpus; do
+        local minikube_key=$(tr _ - <<< "$key")  # e.g., vm-driver
+        local default_var="DEFAULT_MINIKUBE_$(tr a-z A-Z <<< "$key")"  # e.g., DEFAULT_MINIKUBE_VM_DRIVER
```
agianey: I think we're missing defaults for cpu and disk size?

commands/bash/host/config/config
```diff
@@ -12,19 +12,22 @@ function config() {
...
         set)
-            jq -c --arg key "$1" --arg val "$2" '. + { ($key): $val }' "$config_path" > "tmp.$$.json" && mv "tmp.$$.json" "$config_path"
+            jq --arg key "$1" --arg val "$2" 'setpath($key | split("."); $val)' "$config_path" > "tmp.$$.json" && mv "tmp.$$.json" "$config_path"
         ;;
         view)
             cat "$config_path"
         ;;
         unset)
-            jq -c "del(.$1)" "$config_path" > "tmp.$$.json" && mv "tmp.$$.json" "$config_path"
+            jq --arg key "$1" 'delpaths([$key | split(".")])' "$config_path" > "tmp.$$.json" && mv "tmp.$$.json" "$config_path"
```
agianey: some context/reasoning around these changes would be greatly appreciated :)

aladdin.sh
bbil: This is for virtual box right?

This looks like it is replacing the manual steps I used to do whenever creating a new minikube vm (through the virtual box UI).

aladdin.sh
```diff
@@ -86,21 +90,59 @@ function check_and_handle_init() {
...
+
         echo "Starting minikube... (this will take a moment)"
         set_minikube_config
-        minikube start &> /dev/null
+
+        _start_minikube
+
         # Determine if we've installed our bootlocal.sh script to replace the vboxsf mounts with nfs mounts
         if ! "$(minikube ssh -- "test -x /var/lib/boot2docker/bootlocal.sh && echo -n true || echo -n false")"; then
             echo "Installing NFS mounts from host..."
```
bbil: maybe put this "installing..." message in the condition below

commands/bash/host/config/config
ptramsey: Basically, we were passing in the key/value as --args to jq during set, but not during get or unset.  As a result, we got different behavior between those two cases.  In set, a dotted key (e.g., minikube.vm_driver) would be set literally --- `"minikube.vm_driver": "virtualbox"` --- but in unset and get, it would look for 
```
"minikube": {
    "vm_driver": "virtualbox"
}
```

Unfortunately, using --arg is safer than directly interpolating the key/value into the string (reduces confusion around what is parsing '$'; prevents injection).

So, to recap:

for get/unset, we were interpolating dotted values directly into the query string, and getting e.g. .minikube.vm_driver.

For set, we were passing in the key as an argument, but then using it wrong (effectively doing .["minikube.vm_driver"]).

This new code always passes values in as arguments, and then uses them correctly from within jq to address their corresponding locations in the hierarchy.

aladdin.sh
ptramsey: Yes.  I'm not specifying them here, and I explain why in the README --- there's no reason to override the default behavior in minikube.

aladdin.sh
ptramsey: yup!

agianey:
Did some smoke testing on mac, and things looked good

PR #37, opened 2020-03-09
[ALDN-147] Only check AWS credentials when performing cluster operations

jcwilson:
We don't need to validate AWS credentials when working in minikube (ie. LOCAL development). One may not have even defined their `~/.aws/credentials` or `~/.aws/config` files yet, but still want to
try out some local behavior.

[ALDN-147](https://fivestars.atlassian.net/browse/ALDN-147)


jcwilson:
I was able to validate this change on my system by building a custom aladdin image and updating my `aladdin-config/config.json` to use that image instead of the official published one. (thanks for the tip, @ptramsey)

PR #38, opened 2020-03-11
Add assume role capabilities when bastion account is enabled

agianey:
High level:
- There will require an accompanying aladdin-config PR, that specifies which bastion account to use, and different roles we can assume (role that has sufficient privileges to run all aladdin command).
- Then, during aladdin's startup, we will now configure the ~/.aws/config and ~/.aws/credentials files to have the temporary credentials that are issued when calling `aws sts assume-role` for a given profile. 
- Users will now only be required to have aws config/creds file for the bastion account

### DIFF
--- a/aladdin-container.sh
+++ b/aladdin-container.sh
@@ -17,7 +17,7 @@ export SCRIPT_DIR
 export ALADDIN_CONFIG_DIR
 export PY_MAIN
 
-source "$SCRIPT_DIR/shared.sh"
+source "$SCRIPT_DIR/shared.sh" # to load _extract_cluster_config_value
 
 # Test user's aws configuration
 function _test_aws_config() {
@@ -30,7 +30,7 @@ function _test_aws_config() {
         exit 1
     fi
     # Do a test aws cli call for the current aws profile
-    if ! aws s3 ls --profile "$profile" &>/dev/null; then
+    if ! aws iam list-users --profile "$profile" &>/dev/null; then
         echo "Your aws $profile credentials or config may be malformed; please check your ~/.aws/config and ~/.aws/credentials files"
         exit 1
     fi
--- a/aladdin-container.sh
+++ b/aladdin-container.sh
@@ -17,17 +17,21 @@ export SCRIPT_DIR
 export ALADDIN_CONFIG_DIR
 export PY_MAIN
 
+source "$SCRIPT_DIR/shared.sh"
+
 # Test user's aws configuration
-function test_aws_config() {
+function _test_aws_config() {
+    # Test aws configuration for a given profile: $1
     echo "Testing aws configuration..."
+    local profile=$1
     # See if we the current aws profile configured
-    if ! aws configure list --profile "$AWS_PROFILE" &>/dev/null; then
-        echo "Could not find aws profile: $AWS_PROFILE; please check your ~/.aws/config and ~/.aws/credentials files"
+    if ! aws configure list --profile "$profile" &>/dev/null; then
+        echo "Could not find aws profile: $profile; please check your ~/.aws/config and ~/.aws/credentials files"
         exit 1
     fi
     # Do a test aws cli call for the current aws profile
-    if ! aws s3 ls --profile "$AWS_PROFILE" &>/dev/null; then
-        echo "Your aws $AWS_PROFILE credentials or config may be malformed; please check your ~/.aws/config and ~/.aws/credentials files"
+    if ! aws s3 ls --profile "$profile" &>/dev/null; then
+        echo "Your aws $profile credentials or config may be malformed; please check your ~/.aws/config and ~/.aws/credentials files"
         exit 1
     fi
     echo "aws configuration check successful!"
@@ -85,13 +89,7 @@ function environment_init() {
     echo "CLUSTER_CODE = $CLUSTER_CODE"
     echo "NAMESPACE = $NAMESPACE"
 
-    # Kops uses AWS_PROFILE instead of AWS_DEFAULT_PROFILE
-    export AWS_PROFILE="$AWS_DEFAULT_PROFILE"
-
-    # Sanity check the user's aws configuration if init is set
-    if "$INIT" && ! "$IS_LOCAL"; then
-        test_aws_config
-    fi
+    _handle_aws_config
 
     # Make sure we are on local or that cluster has been created before initializing helm, creating namespaces, etc
     if "$IS_LOCAL" || ( kops export kubecfg --name $CLUSTER_NAME &> /dev/null && \
@@ -108,7 +106,7 @@ function environment_init() {
         kubectl config set-context "$NAMESPACE.$CLUSTER_NAME" --cluster "$CLUSTER_NAME" --namespace="$NAMESPACE" --user "$CLUSTER_NAME"
         kubectl config use-context "$NAMESPACE.$CLUSTER_NAME"
 
-        add_and_set_authentication_users
+        _handle_authentication_config
 
         if $INIT; then
             kubectl create namespace --cluster $CLUSTER_NAME $NAMESPACE || true
@@ -123,7 +121,8 @@ function environment_init() {
 }
 
 function _initialize_helm() {
-    if $RBAC_ENABLED; then
+    local rbac_enabled="$(_extract_cluster_config_value rbac_enabled)"
+    if $rbac_enabled; then
         kubectl -n kube-system create serviceaccount tiller || true
         kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount=kube-system:tiller || true
         helm init --service-account=tiller --upgrade --wait || true
@@ -132,9 +131,17 @@ function _initialize_helm() {
     fi
 }
 
-function add_and_set_authentication_users() {
-    # If authentication is enabled, add the appropriate users to our kubeconfig
+function _handle_authentication_config() {
+    # This function adds appropriate users to kubeconfig, and exports necessary AUTHENTICATION variables
+    # export AUTHENTICATION_ENABLED for change-permissions and bash command
+    export AUTHENTICATION_ENABLED="$(_extract_cluster_config_value authentication_enabled)"
     if $AUTHENTICATION_ENABLED; then
+        AUTHENTICATION_ROLES="$(_extract_cluster_config_value authentication_roles)"
+        AUTHENTICATION_ALADDIN_ROLE="$(_extract_cluster_config_value authentication_aladdin_role)"
+        # export AUTHENTICATION_DEFAULT_ROLE for bash command
+        export AUTHENTICATION_DEFAULT_ROLE="$(_extract_cluster_config_value authentication_default_role)"
+        # export AUTHENTICATION_ALLOWED_CHANGE_ROLES for change-permissions command
+        export AUTHENTICATION_ALLOWED_CHANGE_ROLES="$(_extract_cluster_config_value authentication_allowed_change_roles)"
         jq -r '.|keys[]' <<< "$AUTHENTICATION_ROLES" | while read name ; do
             role_arn="$(jq -r --arg name "$name" '.[$name]' <<< $AUTHENTICATION_ROLES)"
             _add_authentication_user_to_kubeconfig "$name" "$role_arn"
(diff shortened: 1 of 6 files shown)


aladdin-container.sh
```diff
@@ -17,17 +17,21 @@ export SCRIPT_DIR
 export ALADDIN_CONFIG_DIR
 export PY_MAIN
 
+source "$SCRIPT_DIR/shared.sh"
```
agianey: Moved the _extract_cluster_config_value function there and source it here

aladdin.sh
```diff
@@ -28,6 +28,8 @@ PATH="$ALADDIN_BIN":"$PATH"
 DEFAULT_MINIKUBE_VM_DRIVER="virtualbox"
 DEFAULT_MINIKUBE_MEMORY=4096
 
+source "$SCRIPT_DIR/shared.sh"
```
agianey: Moved the _extract_cluster_config_value function there and source it here

aladdin-container.sh
```diff
@@ -17,17 +17,21 @@ export SCRIPT_DIR
...
+
 # Test user's aws configuration
-function test_aws_config() {
```
agianey: This function takes in a profile: $1 now

aladdin-container.sh
```diff
@@ -85,13 +89,7 @@ function environment_init() {
     echo "CLUSTER_CODE = $CLUSTER_CODE"
     echo "NAMESPACE = $NAMESPACE"
 
-    # Kops uses AWS_PROFILE instead of AWS_DEFAULT_PROFILE
-    export AWS_PROFILE="$AWS_DEFAULT_PROFILE"
```
agianey: All this removed logic is in _handle_aws_config now

aladdin-container.sh
```diff
@@ -123,7 +121,8 @@ function environment_init() {
 }
 
 function _initialize_helm() {
-    if $RBAC_ENABLED; then
+    local rbac_enabled="$(_extract_cluster_config_value rbac_enabled)"
```
agianey: this config value was only used here, so just extract at use time

aladdin-container.sh
```diff
@@ -132,9 +131,17 @@ function _initialize_helm() {
     fi
 }
 
-function add_and_set_authentication_users() {
-    # If authentication is enabled, add the appropriate users to our kubeconfig
+function _handle_authentication_config() {
+    # This function adds appropriate users to kubeconfig, and exports necessary AUTHENTICATION variables
+    # export AUTHENTICATION_ENABLED for change-permissions and bash command
+    export AUTHENTICATION_ENABLED="$(_extract_cluster_config_value authentication_enabled)"
```
agianey: these variables were previous extracted in aladdin.sh and then passed into the docker run command. now they're just loaded at use time, and exported if required

aladdin-container.sh
```diff
@@ -161,6 +168,61 @@ function _add_authentication_user_to_kubeconfig() {
 EOT
 }
 
+function _handle_aws_config() {
```
agianey: this function got kind of hairy - any feedback welcome, especially on variable naming (for all the values pulled from config)

aladdin-container.sh
```diff
@@ -161,6 +168,61 @@ function _add_authentication_user_to_kubeconfig() {
...
+    # 5) Export AWS_PROFILE=$AWS_DEFAULT_PROFILE because kops uses the former
+
+    # Move aws credentials away from mount so we don't edit the host's aws files
+    cp -r /root/tmp/.aws /root/.aws
+    # See if bastion account is enabled
+    export BASTION_ACCOUNT_ENABLED="$(_extract_cluster_config_value bastion_account_enabled)"
+    if "$BASTION_ACCOUNT_ENABLED"; then
+        # Export BASTION_ACCOUNT_PROFILE, needed by add-aws-assume-role-config command
+        export BASTION_ACCOUNT_PROFILE="$(_extract_cluster_config_value bastion_account_profile)"
+        # Need to unset AWS_DEFAULT_PROFILE because aws requires that entry to exist even if we
```
agianey: This was an annoying issue. see the comment by billwanjohi here for more info https://stackoverflow.com/questions/34134879/aws-the-config-profile-myname-could-not-be-found

aladdin.sh
```diff
@@ -204,21 +197,6 @@ function set_cluster_helper_vars() {
...
-
-    AUTHENTICATION_ENABLED="$(_extract_cluster_config_value authentication_enabled)"
-    if [[ -z "$AUTHENTICATION_ENABLED" ]]; then
-        AUTHENTICATION_ENABLED=false
-    else
-        AUTHENTICATION_ROLES="$(_extract_cluster_config_value authentication_roles)"
-        AUTHENTICATION_ALADDIN_ROLE="$(_extract_cluster_config_value authentication_aladdin_role)"
-        AUTHENTICATION_DEFAULT_ROLE="$(_extract_cluster_config_value authentication_default_role)"
-        AUTHENTICATION_ALLOWED_CHANGE_ROLES="$(_extract_cluster_config_value authentication_allowed_change_roles)"
-    fi
```
agianey: All these are now extracted in aladdin-container.sh where they're used

aladdin.sh
```diff
@@ -336,15 +314,11 @@ function enter_docker_container() {
         -e "IS_PROD=$IS_PROD" \
         -e "IS_TESTING=$IS_TESTING" \
         -e "SKIP_PROMPTS=$SKIP_PROMPTS" \
-        -e "RBAC_ENABLED=$RBAC_ENABLED" \
-        -e "AUTHENTICATION_ENABLED=$AUTHENTICATION_ENABLED" \
-        -e "AUTHENTICATION_ROLES=$AUTHENTICATION_ROLES" \
-        -e "AUTHENTICATION_DEFAULT_ROLE=$AUTHENTICATION_DEFAULT_ROLE" \
-        -e "AUTHENTICATION_ALADDIN_ROLE=$AUTHENTICATION_ALADDIN_ROLE" \
-        -e "AUTHENTICATION_ALLOWED_CHANGE_ROLES=$AUTHENTICATION_ALLOWED_CHANGE_ROLES" \
```
agianey: All these are extracted in aladdin-container now, where they're used

commands/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
```diff
@@ -0,0 +1,67 @@
...
+    else
+
+        payload="$(aws --profile $BASTION_ACCOUNT_PROFILE sts assume-role  \
+            --role-arn $role --duration-seconds "$duration" --role-session-name "$session_name")"
+    fi
+
+    aws configure --profile "$profile_name" set aws_access_key_id "$(jq -r .Credentials.AccessKeyId <<< $payload)"
+    aws configure --profile "$profile_name" set aws_secret_access_key "$(jq -r .Credentials.SecretAccessKey <<< $payload)"
+    aws configure --profile "$profile_name" set aws_session_token "$(jq -r .Credentials.SessionToken <<< $payload)"
+    aws configure --profile "$profile_name" set region "$AWS_DEFAULT_REGION"
```
agianey: Originally, this was echoing export commands, that were later `eval`'d, but that didn't mesh will with how we're using boto in the python code, or kops

commands/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
```diff
@@ -0,0 +1,67 @@
...
+
+        payload="$(aws --profile $BASTION_ACCOUNT_PROFILE sts assume-role  \
+            --role-arn $role --duration-seconds $duration --serial-number $mfa_serial \
+            --token $token --role-session-name $session_name)"
+    else
+
+        payload="$(aws --profile $BASTION_ACCOUNT_PROFILE sts assume-role  \
+            --role-arn $role --duration-seconds "$duration" --role-session-name "$session_name")"
+    fi
+
```
abandali: how about a small check here to ensure `payload` is loaded, before executing the aws/jq commands below, which all _assume_ payload is valid

aladdin.sh
abandali: nit: add a comment at the end of the source command `# to load _extract_cluster_config_value`

aladdin-container.sh
abandali: haha, yeah, it looks ugly but once i read through it all, it makes sense. It's unfortunate but necessary. The names all make sense to me

aladdin-container.sh
```diff
@@ -161,6 +168,61 @@ function _add_authentication_user_to_kubeconfig() {
...
+        # Test aws config for bastion account if INIT and not IS_LOCAL
+        if "$INIT" && ! "$IS_LOCAL"; then
+            _test_aws_config "$BASTION_ACCOUNT_PROFILE"
+        fi
+        # Alias the add assume role config command
+        add_assume_role_config="$ALADDIN_DIR/commands/bash/container/add-aws-assume-role-config/add-aws-assume-role-config"
+        # Need to add aws configuration based on publish configuration.
+        # We need to do this because the publish ECR may be a different aws account than the one
+        # your cluster is provisioned in
+        publish_config="$(_extract_cluster_config_value publish)"
```
jarojasm95: It seems in some places we use the following config pattern:
`prefix_setting_name` (ex. bastion_account_profile)
while in others we put it inside a dictionary and extract using `jq`.

Might be good to unify on a single pattern at some point, not specifically related to this PR though

commands/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+function add_aws_assume_role_config {
```
jarojasm95: It seems this function uses a mix of parameters and env variables to operate, how are we choosing which go where?
My only concern is that reusing functions that do this kind of mix can get hairy

scripts/shared.sh
```diff
@@ -0,0 +1,14 @@
...
+set -eu -o pipefail
+
+function _extract_cluster_config_value() {
+    # Try extracting config from cluster config.json, default config.json, then aladdin config.json
+    local value
+    value="$1"
+    jq -nr --arg value "$value" 'first(inputs | (if .[$value] == null then empty else .[$value] end))' \
+        "$ALADDIN_CONFIG_DIR/$CLUSTER_CODE/config.json" "$ALADDIN_CONFIG_DIR/default/config.json" \
+        "$ALADDIN_CONFIG_DIR/config.json"
+}
```
jarojasm95: nice!

commands/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
agianey: BASTION_ACCOUNT_PROFILE and BASTION_ACCOUNT_ENABLED will always be constant within an invocation of aladdin. All parameters may be different in a single invocation of aladdin (e.g. publish role/publish_profile/publish_mfa_enabled could be distinct from aws_role/aws_profile/aws_mfa_enabled - see the _handle_aws_config function) duration is also a parameter in case someone wants to call it from bash to re-up their creds. 

AWS_DEFAULT_REGION should also never change

commands/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
agianey: with the `set -eu -o pipefail` at the top, it'll immediately error out if the aws command fails, and then aladdin will exit and display that error.

aladdin-container.sh
agianey: Yeah, I was frustrated with the configuration extraction in bash. The reason some are inner nested and some aren't is due to the cluster override configs, specifically that, currently with the way the _extract_cluster_config function is written, they only override at the top level. So if we wanted to override just one part of the inner dictionary in a cluster specific config.json, we currently need to move that entire inner dictionary to the top level.

I can make a ticket to clean up config extraction in the bash part of aladdin

aladdin-container.sh
```diff
@@ -17,17 +17,21 @@ export SCRIPT_DIR
...
-    if ! aws configure list --profile "$AWS_PROFILE" &>/dev/null; then
-        echo "Could not find aws profile: $AWS_PROFILE; please check your ~/.aws/config and ~/.aws/credentials files"
+    if ! aws configure list --profile "$profile" &>/dev/null; then
+        echo "Could not find aws profile: $profile; please check your ~/.aws/config and ~/.aws/credentials files"
         exit 1
     fi
     # Do a test aws cli call for the current aws profile
-    if ! aws s3 ls --profile "$AWS_PROFILE" &>/dev/null; then
-        echo "Your aws $AWS_PROFILE credentials or config may be malformed; please check your ~/.aws/config and ~/.aws/credentials files"
+    if ! aws iam list-users --profile "$profile" &>/dev/null; then
```
agianey: Test iam rather than s3 because, in the least privilege world, a user is not guaranteed to have access to s3. iam is guaranteed because they need list-users to manage their MFA device

jarojasm95: Might be a good chance to clean this up a bit, we know this can fail if minikube time is skewed, maybe show the error output to the user for better debugging?

Could be a separate PR / ticket

aladdin-container.sh
jarojasm95: got it, makes sense, merging nested configs does sound hairy

aladdin.sh
agianey: done

aladdin-container.sh
agianey: updated to show stderr

agianey:
Little bit of scope creep, but made sure to always show stderr, i.e. replacing `&>`'s with `>`'s

aladdin-container.sh
```diff
@@ -17,7 +17,7 @@ export SCRIPT_DIR
 export ALADDIN_CONFIG_DIR
 export PY_MAIN
 
-source "$SCRIPT_DIR/shared.sh"
+source "$SCRIPT_DIR/shared.sh" # to load _extract_cluster_config_value
```
abandali: 👍

aladdin.sh
```diff
@@ -28,7 +28,7 @@ PATH="$ALADDIN_BIN":"$PATH"
 DEFAULT_MINIKUBE_VM_DRIVER="virtualbox"
 DEFAULT_MINIKUBE_MEMORY=4096
 
-source "$SCRIPT_DIR/shared.sh"
+source "$SCRIPT_DIR/shared.sh" # to load _extract_cluster_config_value
```
abandali: 👍

aladdin-container.sh
```diff
@@ -17,17 +17,21 @@ export SCRIPT_DIR
...
-    if ! aws configure list --profile "$AWS_PROFILE" &>/dev/null; then
-        echo "Could not find aws profile: $AWS_PROFILE; please check your ~/.aws/config and ~/.aws/credentials files"
+    if ! aws configure list --profile "$profile" >/dev/null; then
+        echo "Could not find aws profile: $profile; please check your ~/.aws/config and ~/.aws/credentials files"
         exit 1
     fi
     # Do a test aws cli call for the current aws profile
-    if ! aws s3 ls --profile "$AWS_PROFILE" &>/dev/null; then
-        echo "Your aws $AWS_PROFILE credentials or config may be malformed; please check your ~/.aws/config and ~/.aws/credentials files"
+    if ! aws sts get-caller-identity --profile "$profile" >/dev/null; then
```
agianey: On further thought `aws sts get-caller-identity` is even better because it doesn't rely on any permissions, i.e. a user with just access keys and no perms will return something successful with this

aladdin-container.sh
```diff
@@ -161,6 +168,59 @@ function _add_authentication_user_to_kubeconfig() {
...
+    export BASTION_ACCOUNT_ENABLED="$(_extract_cluster_config_value bastion_account_enabled)"
+    if "$BASTION_ACCOUNT_ENABLED"; then
+        # Export BASTION_ACCOUNT_PROFILE, needed by add-aws-assume-role-config command
+        export BASTION_ACCOUNT_PROFILE="$(_extract_cluster_config_value bastion_account_profile)"
+        # Need to unset AWS_DEFAULT_PROFILE because aws requires that entry to exist even if we
+        # are specifying --profile to a separate present entry, which is expected in the bastion case
+        local aws_default_profile_temp="$AWS_DEFAULT_PROFILE"
+        unset AWS_DEFAULT_PROFILE
+        # Test aws config for bastion account
+        "$INIT" && _test_aws_config "$BASTION_ACCOUNT_PROFILE"
```
agianey: This kind of reverts Josh's PR here: https://github.com/fivestars-os/aladdin/pull/37 for when bastion account is enabled, but when disabled it will not check on local. I think it's good to check on local when the bastion account stuff is enabled because now we are explicitly doing aws operations (sts assume-role) in aladdin's initialization, which wasn't a given before.

aladdin-container.sh
agianey: https://fivestars.atlassian.net/projects/ALDN/issues/ALDN-148

aladdin-container.sh
```diff
@@ -85,17 +89,11 @@ function environment_init() {
...
-    if "$INIT" && ! "$IS_LOCAL"; then
-        test_aws_config
-    fi
+    _handle_aws_config
 
     # Make sure we are on local or that cluster has been created before initializing helm, creating namespaces, etc
-    if "$IS_LOCAL" || ( kops export kubecfg --name $CLUSTER_NAME &> /dev/null && \
-        kops validate cluster --name $CLUSTER_NAME ) &> /dev/null; then
+    if "$IS_LOCAL" || ( kops export kubecfg --name $CLUSTER_NAME > /dev/null && \
+        kops validate cluster --name $CLUSTER_NAME ) > /dev/null; then
```
jarojasm95: While looking into a Jenkins Cluster issue I found myself unable to launch `aladdin bash`, reason being is that `kops validate cluster` was failing (the cluster went down, yikes), and so `_handle_authentication_config` was not being called (below) and so aladdin was exiting with:
```
/root/aladdin/commands/bash/container/bash/bash: line 12: AUTHENTICATION_ENABLED: unbound variable
```

PR #42, opened 2020-04-22
[ALDN-166] Sync date and time automatically every time

phani-kandula-fivestars:
[ALDN-166](https://fivestars.atlassian.net/browse/ALDN-166)

### DIFF
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -284,6 +284,14 @@ function prepare_docker_subcommands() {
     fi
 }
 
+function synchronize_datetime()
+{
+  # When minikube wakes up from sleep, date or time could be out of sync.
+  # Take date + time from host and set it on minikube.
+  echo "Synchronizing date and time on minikube with the host"
+  ssh -i ~/.minikube/machines/minikube/id_rsa docker@$(minikube ip) "docker run --rm --privileged --pid=host alpine nsenter -t 1 -m -u -n -i date -u $(date -u +%m%d%H%M%Y)"
+}
+
 function enter_docker_container() {
     if "$IS_PROD" && ! "$SKIP_PROMPTS"; then
         confirm_production
@@ -378,4 +386,5 @@ copy_ssh_to_minikube
 set_cluster_helper_vars
 handle_ostypes
 prepare_docker_subcommands
+synchronize_datetime
 enter_docker_container "$@"
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -288,8 +288,10 @@ function synchronize_datetime()
 {
   # When minikube wakes up from sleep, date or time could be out of sync.
   # Take date + time from host and set it on minikube.
-  echo "Synchronizing date and time on minikube with the host"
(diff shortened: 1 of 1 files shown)


aladdin.sh
```diff
@@ -378,4 +386,5 @@ copy_ssh_to_minikube
 set_cluster_helper_vars
 handle_ostypes
 prepare_docker_subcommands
+synchronize_datetime
```
jarojasm95, line 389: I'm a bit worried on performance, do we need to run this every time?

jarojasm95: can we detect time skew and then run this based on that?

aladdin.sh
```diff
@@ -284,6 +284,14 @@ function prepare_docker_subcommands() {
     fi
 }
 
+function synchronize_datetime()
+{
+  # When minikube wakes up from sleep, date or time could be out of sync.
+  # Take date + time from host and set it on minikube.
+  echo "Synchronizing date and time on minikube with the host"
+  ssh -i ~/.minikube/machines/minikube/id_rsa docker@$(minikube ip) "docker run --rm --privileged --pid=host alpine nsenter -t 1 -m -u -n -i date -u $(date -u +%m%d%H%M%Y)"
```
jarojasm95, line 292: I think this might need some path normalizing depending on the OS

jarojasm95: actually, I think using `minikube ssh` might be better than plain `ssh`

agianey: we can run the `docker run` command by itself from within aladdin-container.sh

aladdin.sh
jarojasm95: I think we can detect time skew by putting the host time on an env variable when starting the container, then check the diff between that and the aladdin container time, if that is greater than X threshold then run the synchronize

aladdin.sh
ptramsey, line 292: This may be a silly question, but is there a reason not to just use `minikube ssh`?

ptramsey, line 292: Also, note that --vm-driver=none is a supported configuration for minikube in aladdin.  In this configuration, there is no minikube vm, and this command will fail.  You should check 
```
if test $(minikube config get vm-driver) != "none"; then
...
```

ptramsey: Could this command be replaced by,
```
minikube ssh "sudo date -s @$(date +%s)"
```
?

phani-kandula-fivestars: As per PTR's suggestion: skipped docker step. Just minikube ssh and set date.

phani-kandula-fivestars: Yes. Changed it now.

phani-kandula-fivestars: Thanks. Added that check.

aladdin.sh
ptramsey: How concerned are we about performance?  I ran this 10 times and it looked like it was averaging about 3/4 of a second.  Is that worth optimizing to avoid?  In particular, I don't imagine we can detect clock skew without sshing into minikube, which is what takes the majority of that time (just setting the date from inside the minikube vm takes 0.001s)

jcwilson: no love for the "do on init" suggestion?

PR #43, opened 2020-04-22
Pass extra args through to build command

jcwilson:
This change lets one provide extra arguments to the build script specified
in the `lamp.json` file. This can be useful during development if your build
takes a relatively long time and you want to only perform a partial build.

### DIFF
--- a/commands/python/command/build.py
+++ b/commands/python/command/build.py
@@ -3,15 +3,17 @@ from project.project_conf import ProjectConf
 
 
 def parse_args(sub_parser):
-    subparser = sub_parser.add_parser('build',
-                                      help='Build a project\'s docker images for local development')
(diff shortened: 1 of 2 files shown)


commands/python/command/build.py
```diff
@@ -3,15 +3,17 @@
 
 
 def parse_args(sub_parser):
-    subparser = sub_parser.add_parser('build',
-                                      help='Build a project\'s docker images for local development')
+    subparser = sub_parser.add_parser(
+        "build", help="Build a project's docker images for local development"
+    )
+    subparser.add_argument("build_args", nargs="*")
```
jcwilson, line 9: should we use `argparse.REMAINDER` like `cmd.py` does instead?

PR #44, opened 2020-04-22
Add support for commands-base:2.0.0

jcwilson:
The `commands-base:2.0.0` image moves the `command.py` script to a new location and name:
`/usr/local/bin/aladdin_command`. It also assumes responsibility for ensuring that it is executable so
aladdin can just invoke the script or program. This is to allow developers to mount their host files
into their command containers at `/code` without stomping on this necessary script.

Projects that wish to move to the new `commands-base:2.0.0` image should update their commands'
Dockerfiles to derive from this image. Aladdin will check for the existence of an executable named
`aladdin_command` on the `PATH` before falling back to the previous behavior.

If one wishes to use an alternative implementation of the command script provided by
`commands-base:2.0.0`, one only needs to create an executable named `aladdin_command` on their
image's `PATH` that accepts the same parameter's as the default implementation in
`commands-base:2.0.0`. (from #26)

See fivestars-os/commands-base#3 for further details.

### DIFF
--- a/commands/python/command/cmd.py
+++ b/commands/python/command/cmd.py
@@ -3,17 +3,27 @@ import logging
 import subprocess
 import sys
 
+from jmespath import search
+
 from arg_tools import add_namespace_argument
 from libs.k8s.kubernetes import Kubernetes
+from project.project_conf import ProjectConf
 
 
 def parse_args(sub_parser):
-    subparser = sub_parser.add_parser('cmd',
-                                      help='Run commands on specifc project via its commands pod')
-    subparser.add_argument('app_name', type=str, metavar='app_name',
-                           help='app to run commands against')
-    subparser.add_argument('command_args', type=str, metavar='comm_args',
-                           help='command and its args', nargs=argparse.REMAINDER)
+    subparser = sub_parser.add_parser(
+        "cmd", help="Run commands on specifc project via its commands pod"
+    )
+    subparser.add_argument(
+        "app_name", type=str, metavar="app_name", help="app to run commands against"
+    )
+    subparser.add_argument(
+        "command_args",
+        type=str,
+        metavar="comm_args",
+        help="command and its args",
+        nargs=argparse.REMAINDER,
+    )
     add_namespace_argument(subparser)
     subparser.set_defaults(func=cmd_args)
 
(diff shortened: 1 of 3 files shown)


agianey:
This requires that we now be in the correct project's directory when calling the cmd command right? i.e. to inspect the lamp.json file. We've generally avoided this in the past for "non local" commands, although i think that is not true for some of our internal plugins.

I'm ok with changing this requirement, but if there's a way to do it without inspecting the lamp.json that may be preferable.

jcwilson:
> This requires that we now be in the correct project's directory when calling the cmd command right? i.e. to inspect the lamp.json file. We've generally avoided this in the past for "non local" commands, although i think that is not true for some of our internal plugins.

good point. I think the lamp file versioning isn't going to work for these cases. It's not difficult to add a check with an initial `kub_exec()` call to check for the existence of the new file before attempting to exec it. Want me to give that a shot instead?

commands/python/libs/k8s/kubernetes.py
```diff
@@ -65,8 +65,8 @@ def kub_exec(self, pod_name, container_name, *command, return_output=False, term
             cmd_list = self._kub_cmd('exec', flags, pod_name, '--', *command)
 
         if return_output:
-            DEVNULL = open(os.devnull, 'w')
-            return subprocess.check_output(cmd_list, stderr=DEVNULL)
+            with open(os.devnull, "w") as devnull:
+                return subprocess.check_output(cmd_list, stderr=devnull)
```
jcwilson, line 69: Figured I'd fix this while I was in here

commands/python/command/cmd.py
```diff
@@ -24,13 +31,38 @@ def cmd_args(args):
...
-                                return_output=True).decode(sys.stdout.encoding).strip()
+        # In commands-base:2.0.0, we moved the commands.py script to /usr/local/bin/aladdin_cmd_v2
+        # and made it executable. It's expected that the implementer will provide a she-bang line to
+        # indicate which interpreter to use (or none at all if they are using a compiled executable)
+        k.kub_exec(
+            pod_name,
+            f"{app_name}-commands",
+            "/bin/bash",
+            "-c",
+            "test -x /usr/local/bin/aladdin_cmd_v2",
```
jcwilson, line 46: I updated this to `/usr/local/bin/aladdin_cmd_v2` to future-proof us for potential future version changes. but maybe we leave it as `/usr/local/bin/aladdin_cmd` and deal with any future version changes by adding a version suffix in the future. I don't know how much more generic we can make this, so i don't anticipate many interface updates in the future. Maybe we start injecting arguments from aladdin as we call the command container?

commands/python/libs/k8s/kubernetes.py
ptramsey: :+1: :+1: :+1:

PR #45, opened 2020-05-04
Improve Startup Performance

jarojasm95:
This PR tries to improve the performance of aladdin commands by decreasing the number of commands run / checks run every time a command is issued and in some places by using a local file cache for data that can be re-used on subsequent runs (such as aws credentials)


aladdin-container.sh
```diff
@@ -91,33 +106,38 @@ function environment_init() {
...
+    if "$IS_LOCAL"; then
+        mkdir -p $HOME/.kube/
+        cp $HOME/.kube_local/config $HOME/.kube/config
+        # Replace e.g. /Users/ptr/.minikube with /root/.minikube in the minikube conf dir path
+        # (since that's where it will be mounted within the aladdin container)
+        sed 's|: \?.*\.minikube|: /root/.minikube| ; /: \/root\/.minikube/ s|\\|/|g' $HOME/.kube_local/config > $HOME/.kube/config
+    elif test -z "$(get_or_set_cache "kops_$CLUSTER_NAME")" || ! test -f "$HOME/.kube_local/$CLUSTER_NAME.config"; then
+        echo "Getting cluster config and verifying cluster using kops"
+        if ( kops export kubecfg --name $CLUSTER_NAME > /dev/null && \
+            kops validate cluster --name $CLUSTER_NAME ) > /dev/null; then
```
jarojasm95, line 118: previously we were doing these kops commands on each command that wasn't on the minikube cluster

aladdin-container.sh
```diff
@@ -205,19 +231,26 @@ function _handle_aws_config() {
...
-        "$add_assume_role_config" "$publish_role" "$publish_profile" "$publish_mfa_enabled" 3600 # 1 hour
         # Need to add aws configuration for current cluster's aws account
-        aws_profile="$(_extract_cluster_config_value bastion_account_profile_to_assume)"
-        aws_role="$(_extract_cluster_config_value bastion_account_role_to_assume)"
-        aws_mfa_enabled="$(_extract_cluster_config_value bastion_account_mfa_enabled)"
-        "$add_assume_role_config" "$aws_role" "$aws_profile" "$aws_mfa_enabled" 3600 # 1 hour
+        bastion_profile="$(_extract_cluster_config_value bastion_account_profile_to_assume)"
+        bastion_role="$(_extract_cluster_config_value bastion_account_role_to_assume)"
+        bastion_mfa_enabled="$(_extract_cluster_config_value bastion_account_mfa_enabled)"
+        if [[ $bastion_role != $publish_role ]]; then
```
jarojasm95, line 238: if the roles are the same then there's no need to get credentials for both

aladdin.sh
```diff
@@ -54,39 +56,43 @@ function get_plugin_dir() {
...
-    mkdir -p "$(dirname "$last_launched_file")" && touch "$last_launched_file"
-    previous_run="$(cat "$last_launched_file")"
-    if [[ "$current_time" -gt "$((${previous_run:-0}+init_every))" || "$previous_run" -gt "$current_time" ]]; then
+    if ! test -f "$HOME/.aladdin/infra/installed"; then
         INIT=true
     fi
-    # Handle initialization logic
-    if "$INIT"; then
+    if $INIT; then
+        clear_cache_file
```
jarojasm95, line 71: on user `--init` nuke the cache file

aladdin.sh
```diff
@@ -54,39 +56,43 @@ function get_plugin_dir() {
...
+
+    # Use host docker if available
+    if docker info > /dev/null 2>/dev/null; then
+        case "$OSTYPE" in
+            darwin*) USING_HOST_DOCKER=true ;;
+        esac
+        if $USING_HOST_DOCKER; then
+            echo "Using host docker engine for aladdin"
+        fi
+    fi
```
jarojasm95, line 93: Maybe a scope creep, but the aladdin container can run on the host docker engine

aladdin.sh
```diff
@@ -54,39 +56,43 @@ function get_plugin_dir() {
...
+    if [[ $CLUSTER_CODE == "minikube" || $USING_HOST_DOCKER == "false" ]]; then
+        check_or_start_minikube
```
jarojasm95, line 95: if we're not acting against minikube and the host machine has docker available, skip minikube

aladdin.sh
```diff
@@ -295,21 +297,28 @@ function enter_docker_container() {
...
 
     case "$OSTYPE" in
         cygwin*) ssh_src="/.ssh" ;;
         *)       ssh_src="$(pathnorm ~/.ssh)" ;;
     esac
 
+    # How to mount docker parts
+    mounts="$mounts -v /var/run/docker.sock:/var/run/docker.sock"
+    if $USING_HOST_DOCKER ; then
+        # Special case mounting host docker on linux/win?
```
jarojasm95, line 310: does this mount need to change for linux/win?

commands/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
```diff
@@ -30,13 +38,15 @@ function add_aws_assume_role_config {
...
-
+            --token $token --role-session-name $(_get_session_name))"
+    elif test -z "$cached_credentials"; then
         payload="$(aws --profile $BASTION_ACCOUNT_PROFILE sts assume-role  \
-            --role-arn $role --duration-seconds "$duration" --role-session-name "$session_name")"
+            --role-arn $role --duration-seconds "$duration" --role-session-name $(_get_session_name))"
+        get_or_set_cache $params $(time_plus_offset $duration) "$payload"
+    else
+        echo "Using cached credentials"
+        payload="$cached_credentials"
```
jarojasm95, line 48: we create credentials valid for an hour, re-use them

scripts/shared.sh
```diff
@@ -4,11 +4,94 @@
...
+    if ! test -z "$value"; then
+        echo "$value"
+        return 0
+    fi
+    value=$(_extract_from_file "$key" "$ALADDIN_CONFIG_DIR/config.json")
+    if ! test -z "$value"; then
+        echo "$value"
+        return 0
+    fi
+}
```
jarojasm95, line 39: improved this function to be able to get nested values across several config files

aladdin-container.sh
```diff
@@ -91,33 +106,38 @@ function environment_init() {
...
+        # (since that's where it will be mounted within the aladdin container)
+        sed 's|: \?.*\.minikube|: /root/.minikube| ; /: \/root\/.minikube/ s|\\|/|g' $HOME/.kube_local/config > $HOME/.kube/config
+    elif test -z "$(get_or_set_cache "kops_$CLUSTER_NAME")" || ! test -f "$HOME/.kube_local/$CLUSTER_NAME.config"; then
+        echo "Getting cluster config and verifying cluster using kops"
+        if ( kops export kubecfg --name $CLUSTER_NAME > /dev/null && \
+            kops validate cluster --name $CLUSTER_NAME ) > /dev/null; then
+
+            _handle_authentication_config true
             cp $HOME/.kube/config $HOME/.kube_local/$CLUSTER_NAME.config
+            get_or_set_cache "kops_$CLUSTER_NAME" $(time_plus_offset 3600*24)
```
jarojasm95, line 122: arbitrary expiration of 24 hrs

aladdin.sh
```diff
@@ -54,39 +56,43 @@ function get_plugin_dir() {
...
-        local aladdin_image="$(jq -r '.aladdin.repo' "$ALADDIN_CONFIG_FILE"):$(jq -r '.aladdin.tag' "$ALADDIN_CONFIG_FILE")"
-        docker pull "$aladdin_image"
-        echo "$current_time" > "$last_launched_file"
     else
         "$SCRIPT_DIR"/infra_k8s_check.sh
-        enter_minikube_env
+    fi
+    if test -z "$(get_or_set_cache "${ALADDIN_IMAGE}")"; then
+        docker pull "$ALADDIN_IMAGE"
+        get_or_set_cache "${ALADDIN_IMAGE}" $(time_plus_offset 3600*24)
```
jarojasm95, line 82: arbitrary expiration of 24 hrs

commands/bash/host/host/host
```diff
@@ -10,9 +11,9 @@ function host {
 		# On mac/linux :
 		# - sudo vim /etc/hosts
 		# and add this:
-		$MINIKUBE_IP minikube
+		$minikube_ip minikube
 	EOF
-    kubectl get ingress ingress -o=json 2>/dev/null | jq '.spec.rules[]?.host' --raw-output | grep -v '^$' | xargs -I % echo "$MINIKUBE_IP %"
+    kubectl get ingress ingress -o=json 2>/dev/null | jq '.spec.rules[]?.host' --raw-output | grep -v '^$' | xargs -I % echo "$minikube_ip %"
```
jarojasm95, line 16: moved this plugin to host, here (and in a test) were the only places where `MINIKUBE_IP` was being used

aladdin.sh
```diff
@@ -220,7 +224,7 @@ function exec_host_plugin() {
 }
 
 function pathnorm(){
-    # Normalize the path, the path should exists
+    # Normalize the path, the path should exist
```
agianey, line 227: shoutout to mickael

aladdin.sh
jarojasm95: good catch!

aladdin-container.sh
```diff
@@ -136,19 +156,25 @@ function _initialize_helm() {
...
+        export AUTHENTICATION_ALADDIN_ROLE="$(_extract_cluster_config_value authentication_aladdin_role)"
         # export AUTHENTICATION_DEFAULT_ROLE for bash command
         export AUTHENTICATION_DEFAULT_ROLE="$(_extract_cluster_config_value authentication_default_role)"
         # export AUTHENTICATION_ALLOWED_CHANGE_ROLES for change-permissions command
         export AUTHENTICATION_ALLOWED_CHANGE_ROLES="$(_extract_cluster_config_value authentication_allowed_change_roles)"
-        jq -r '.|keys[]' <<< "$AUTHENTICATION_ROLES" | while read name ; do
-            role_arn="$(jq -r --arg name "$name" '.[$name]' <<< $AUTHENTICATION_ROLES)"
-            _add_authentication_user_to_kubeconfig "$name" "$role_arn"
-        done
-        kubectl config set-context "$NAMESPACE.$CLUSTER_NAME" --cluster "$CLUSTER_NAME" --namespace="$NAMESPACE" --user "$AUTHENTICATION_ALADDIN_ROLE"
```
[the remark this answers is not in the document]
jarojasm95: agree! the flag was to cover the case where this function is called twice, I've split up the functions and added a comment

aladdin.sh
jarojasm95: I've used it for a while and seems to work fine, but would rather get this released separately so I've removed it from this PR

aladdin.sh
jarojasm95: removed

aladdin.sh
jarojasm95: removed

aladdin-container.sh
```diff
@@ -91,33 +106,37 @@ function environment_init() {
...
+        mkdir -p $HOME/.kube/
+        cp $HOME/.kube_local/config $HOME/.kube/config
+        # Replace e.g. /Users/ptr/.minikube with /root/.minikube in the minikube conf dir path
+        # (since that's where it will be mounted within the aladdin container)
+        sed 's|: \?.*\.minikube|: /root/.minikube| ; /: \/root\/.minikube/ s|\\|/|g' $HOME/.kube_local/config > $HOME/.kube/config
+    elif test -z "$(get_or_set_cache "kops_$CLUSTER_NAME")" || ! test -f "$HOME/.kube_local/$CLUSTER_NAME.config"; then
+        echo "Getting cluster config and verifying cluster using kops"
+        if ( kops export kubecfg --name $CLUSTER_NAME > /dev/null && \
+            kops validate cluster --name $CLUSTER_NAME ) > /dev/null; then
+            _add_authentication_users_to_kubeconfig
```
jarojasm95, line 119: we only need to add the users to kubeconfig when initially exporting from kops

aladdin-container.sh
```diff
@@ -91,33 +106,38 @@ function environment_init() {
...
-        kubectl config use-context "$NAMESPACE.$CLUSTER_NAME"
-
+    else
+        echo "Using cached cluster config"
+        mkdir -p $HOME/.kube/
+        cp $HOME/.kube_local/$CLUSTER_NAME.config $HOME/.kube/config
+    fi
+    # Make sure we are on local or that cluster has been created before initializing helm, creating namespaces, etc
+    if "$IS_LOCAL" || ! test -z "$(get_or_set_cache "kops_$CLUSTER_NAME")"; then
         _handle_authentication_config
```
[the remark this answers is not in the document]
jarojasm95: I've removed that parameter to make this function a bit cleaner

aladdin.sh
jarojasm95: improved it to make sure the image is available

Dockerfile
```diff
@@ -26,18 +26,14 @@ ARG HELM_VERSION=2.16.1
...
-
-RUN go get -u -v sigs.k8s.io/aws-iam-authenticator/cmd/aws-iam-authenticator
+	&& chmod 755 /bin/kubectl \
+    && curl -L -o /bin/kops https://github.com/kubernetes/kops/releases/download/$KOPS_VERSION/kops-linux-amd64 \
+	&& chmod 755 /bin/kops \
+    && curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && cp linux-amd64/helm \
+	/bin/helm && chmod 755 /bin/helm && helm init --client-only \
+    && curl -L -o- https://download.docker.com/linux/static/stable/x86_64/docker-$DOCKER_VERSION.tgz | tar -zxvf - && cp docker/docker \
+    /usr/bin/docker && chmod 755 /usr/bin/docker \
+    && go get -u -v sigs.k8s.io/aws-iam-authenticator/cmd/aws-iam-authenticator
```
ptramsey, line 36: Awesome, these seem like good things to consolidate

README.md
```diff
@@ -168,6 +168,7 @@ We have several aladdin commands used for development and deployment. Note that
 - `--dev` mount host's aladdin directory onto aladdin container. Useful when developing aladdin.
 - `--skip-prompts` skip any confirmation messages during aladdin execution. Useful when automating commands.
 - `--non-terminal` run aladdin container without tty.
+- `--cluster-creating` skips cluster init commands such as namespace creation and helm init, useful for entering `aladdin bash` when the cluster is just being created
```
ptramsey, line 171: Might it make sense instead to detect this?  Having a flag you have to know to pass to jump past certain things at an early stage seems a little awkward, particularly since you're telling aladdin something it already knows (that the cluster is creating).  If we do keep this, I might suggest naming it something that describes what it does (rather than what it's for), so that the caller can get a sense of the implications of using it (maybe `--unsafe`?  `--skip-init`?)

aladdin-container.sh
```diff
@@ -75,13 +75,28 @@ function exec_command_or_plugin() {
 }
 
 function _replace_aws_secret() {
+    if ! test -z "$(get_or_set_cache "aws_secret_${CLUSTER_NAME}")"; then
```
ptramsey, line 78: `! test -z` is slightly redundant; `test -n` would be better.

scripts/shared.sh
```diff
@@ -4,11 +4,94 @@
...
+    done
+    return 1
+}
+
+function clear_cache_file(){
+    rm -f "$HOME/.aladdin/infra/cache.json"
+}
+
+function get_or_set_cache(){
+    local cache_file="$HOME/.aladdin/infra/cache.json"
```
ptramsey, line 62: It would be nice to have some documentation at the top of this function about the ways it can be called and their implications.

scripts/shared.sh
```diff
@@ -4,11 +4,94 @@
...
+    elif test -z "$data"; then
+        # function called like `get_or_set_cache "key" "expiration"`
+        data=true
+    fi
+
+    if [[ "$(date +'%s')" -gt "$expiration" ]]; then
+        # if the key is expired, clear the data
+        contents="$(jq --arg key "${key}" '.[$key] = {}' $cache_file)"
+        echo "${contents}" > $cache_file
+    elif ! test -z "$data"; then
```
ptramsey, line 83: `test -n`

scripts/shared.sh
```diff
@@ -4,11 +4,94 @@
...
+    local cache_file="$HOME/.aladdin/infra/cache.json"
+    local key="$(make_hash $1)"
+    local expiration="${2:-}"
+    local data="${3:-}"
+    local contents
+
+    if ! jq -r . $cache_file &> /dev/null; then
+        echo "{}" > "$cache_file"
+    fi
+    if test -z "$expiration"; then
```
ptramsey, line 71: Rather than a single get_or_set function that does both, for the sake of clarity in the caller, it might be nice to have separate get and set/`set_and_return`.  There's a lot of conditional logic in this function to allow it to wear multiple hats, and I'm not sure that the result improves clarity in the caller.

scripts/shared.sh
ptramsey: Could it not before?

scripts/shared.sh
```diff
@@ -4,11 +4,94 @@
 set -a
 set -eu -o pipefail
 
+function _extract_from_file() {
+    local key="$1"
+    local config_file="$2"
+    local value
+    if [[ -f $config_file ]]; then
+        value="$(eval "jq -r .${key} $config_file")"
```
ptramsey, line 12: why is this eval necessary?

aladdin.sh
```diff
@@ -263,28 +256,26 @@ function handle_ostypes() {
...
+function enter_docker_container() {
+
     # if this is not production or staging, we are mounting kubernetes folder so that
     # config maps and other settings can be customized by developers
+    local mounts=""
     if "$DEV"; then
-        DEV_CMD="-v $(pathnorm "$ALADDIN_DIR"):/root/aladdin"
-        DEV_CMD="$DEV_CMD -v /:/minikube_root" # mount the whole minikube system
-        DEV_CMD="$DEV_CMD --workdir /minikube_root$(pathnorm "$PWD")"
+        mounts="$mounts -v $(pathnorm "$ALADDIN_DIR"):/root/aladdin"
```
ptramsey, line 265: could be good to use an array for this; avoids escaping issues with paths that have spaces in them.  Minor nit though, and would accept it as it is!

aladdin.sh
```diff
@@ -156,6 +151,11 @@ function check_or_start_minikube() {
             check_or_start_minikube
         fi
     fi
+    if ! minikube addons list | grep "ingress " | grep enabled > /dev/null; then
+        minikube addons enable ingress > /dev/null
```
ptramsey, line 155: nice.

aladdin-container.sh
ptramsey, line 78: Also, this would be a place where it might be night to have a separate get_cache.  No need for the reader to wonder what that function is doing here and whether it's changing anything.

aladdin.sh
jarojasm95: My bash-fu is weak, would you mind pointing me to an example?

scripts/shared.sh
jarojasm95: hmm not sure! this is how I got it to work hehe, my bash-fu is admittedly weak, but `jq` was complaining about ${key}

scripts/shared.sh
jarojasm95: before it could only get flat values merged across configs iirc, now it can also get nested values

scripts/shared.sh
jcwilson: I'm also intrigued by the apparent necessity for the `eval`

This looks like you might be trying to support deep key lookup. I think this could all be one-lined to

```
jq -er ".$key // empty" "$config_file"
```
You won't need to check the return `$value` anymore. Furthermore, callers of this function can use the exit status to determine success/failure, so in theory you could rewrite `_extract_cluster_config_value()` to invoke those three queries in short-circuiting fashion (`a || b || c`)

scripts/shared.sh
jcwilson: does anything use this nested lookup yet?

scripts/shared.sh
```diff
@@ -4,11 +4,106 @@
...
+}
+
+function get_cached() {
+    local key="$(make_hash $1)"
+    _get_or_set_cache $key
+}
+
+function _get_or_set_cache(){
+    local cache_file="$HOME/.aladdin/infra/cache.json"
+    local key="$(make_hash $1)"
```
jcwilson, line 75: doesn't this hash the key twice if called from `get_cache` or `set_cache`? I guess that's ok as long as no one tries to use `_get_or_set_cache()` directly.

scripts/shared.sh
```diff
@@ -4,11 +4,106 @@
...
+function clear_cache_file(){
+    rm -f "$HOME/.aladdin/infra/cache.json"
+}
+
+function set_cache() {
+    local key="$(make_hash $1)"
+    local expiration="${2:-}"
+    local data="${3:-}"
+    _get_or_set_cache $key $expiration $data
+}
```
jcwilson, line 66: the arguments seem kind of backwards. generally caching APIs have the TTL as optional. i don't think it would be overly burdensome to require the second argument if the TTL is needed. if we really wanted to mark something as value-less for the sake of checking an expired time (like the aladdin image refresh, i might suggest a separate set of functions like `set_ttl` and `check_ttl` that just take a name and a TTL (when setting)

scripts/shared.sh
jcwilson: I agree.

I also think we could leverage `jq` a bit more to reduce our code here. See the aforementioned `-e` and `// empty` features.

aladdin.sh
ptramsey: SO

The syntax for this would be, e.g.,
```
local -a mounts  # declares $mounts as an array
if "$DEV"; then
    # [@] and [*] are both ways of getting all elements of an array.  Which you use depends on which behaviour you 
    # want when you surround them with quotes --- whether you want the quotes to go around each array element, or 
    # around the whole array.
    # "${mounts[@]}" expands to every element of the array $mounts as a separately quoted string
    #                (ie, (a b "c d") becomes "a" "b" "c d")
    # "${mounts[*])" would be every element concatenated together with spaces
    #                (ie, (a b "c d") becomes "a b c d")
    # Without the surrounding quotes, either version would be treated like a string with spaces in it (ie, broken up into
    # words).  So, for both ${mounts[@]} and ${mounts[*]}:
    #                ((a b "c d") becomes a b c d)
    # [@] is what you want here, as each array element is an argument.

    mounts=("${mounts[@]}" -v $(pathnorm "$ALADDIN_DIR"):/root/aladdin)
fi
....
docker run $FLAGS
....
"${mounts[@]}"
```

aladdin-container.sh
```diff
@@ -91,33 +106,30 @@ function environment_init() {
...
+        kops export kubecfg --name $CLUSTER_NAME > /dev/null
+        _add_authentication_users_to_kubeconfig
+        cp $HOME/.kube/config $HOME/.kube_local/$CLUSTER_NAME.config
+    else
+        echo "Using cached cluster config"
+        mkdir -p $HOME/.kube/
+        cp $HOME/.kube_local/$CLUSTER_NAME.config $HOME/.kube/config
+    fi
+    _handle_authentication_config
+    kubectl config use-context "$NAMESPACE.$CLUSTER_NAME"
```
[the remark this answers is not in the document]
jarojasm95: kubectl will act on `$HOME/.kube`, this is a directory that exists only on the container (not mounted on host or otherwise), so there won't be any crosstalk 👍

aladdin-container.sh
```diff
@@ -134,24 +146,32 @@ function _initialize_helm() {
...
     if $AUTHENTICATION_ENABLED; then
-        AUTHENTICATION_ROLES="$(_extract_cluster_config_value authentication_roles)"
-        AUTHENTICATION_ALADDIN_ROLE="$(_extract_cluster_config_value authentication_aladdin_role)"
+        # export AUTHENTICATION_ALADDIN_ROLE for change_to_aladdin_permission helper command
+        export AUTHENTICATION_ALADDIN_ROLE="$(_extract_cluster_config_value authentication_aladdin_role)"
         # export AUTHENTICATION_DEFAULT_ROLE for bash command
         export AUTHENTICATION_DEFAULT_ROLE="$(_extract_cluster_config_value authentication_default_role)"
         # export AUTHENTICATION_ALLOWED_CHANGE_ROLES for change-permissions command
         export AUTHENTICATION_ALLOWED_CHANGE_ROLES="$(_extract_cluster_config_value authentication_allowed_change_roles)"
-        jq -r '.|keys[]' <<< "$AUTHENTICATION_ROLES" | while read name ; do
```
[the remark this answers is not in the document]
jarojasm95: yep 👍

aladdin-container.sh
jarojasm95: did it 👍

scripts/shared.sh
jarojasm95: I've cleaned this up, good catch

scripts/shared.sh
jarojasm95: I simplified this function and cleaned up how it's called to be a standard `set_cache <key> <data> <optional ttl>`

scripts/shared.sh
jarojasm95: not yet

scripts/shared.sh
jarojasm95: so using `jq -er ".$key // empty" "$config_file"` did not work because `false` values would end up as `empty`, I did however clean this up a bunch and removed the use of `eval`

scripts/shared.sh
jarojasm95: fixed 👍

scripts/shared.sh
ptramsey: :+1: :+1: :+1: 
Thanks!

aladdin.sh
```diff
@@ -263,28 +247,26 @@ function handle_ostypes() {
...
+function enter_docker_container() {
+
     # if this is not production or staging, we are mounting kubernetes folder so that
     # config maps and other settings can be customized by developers
+    local -a mounts
     if "$DEV"; then
-        DEV_CMD="-v $(pathnorm "$ALADDIN_DIR"):/root/aladdin"
-        DEV_CMD="$DEV_CMD -v /:/minikube_root" # mount the whole minikube system
-        DEV_CMD="$DEV_CMD --workdir /minikube_root$(pathnorm "$PWD")"
+        mounts=("${mounts[@]}" "-v $(pathnorm "$ALADDIN_DIR"):/root/aladdin")
```
ptramsey, line 256: I'm not sure this is quite right.  You don't want `"-v $(pathnorm "$ALADDIN_DIR"):/root/aladdin"` to be a single argument, I think --- I think you want "-v" and "$(pathnorm "$ALADDIN_DIR"):/root/aladdin` to be separate arguments.

aladdin.sh
```diff
@@ -263,28 +247,26 @@ function handle_ostypes() {
...
-    # If on minikube mount minikube credentials
     if "$IS_LOCAL"; then
-        MINIKUBE_CMD="-v $(pathnorm ~/.minikube):/root/.minikube"
-        MINIKUBE_CMD="$MINIKUBE_CMD -v /:/minikube_root" # mount the whole minikube system
-        MINIKUBE_CMD="$MINIKUBE_CMD --workdir /minikube_root$(pathnorm "$PWD")"
+        mounts=("${mounts[@]}" "-v $(pathnorm ~/.minikube):/root/.minikube")
+    fi
+    if "$IS_LOCAL" || "$DEV"; then
+        mounts=("${mounts[@]}" "-v /:/aladdin_root")
+        mounts=("${mounts[@]}" "--workdir /aladdin_root$(pathnorm "$PWD")")
```
ptramsey, line 263: Same here.  If you wouldn't put quotes around it on the command line, you shouldn't here.

aladdin.sh
```diff
@@ -295,25 +277,27 @@ function enter_docker_container() {
...
-    local aladdin_image="${IMAGE:-"$(jq -r '.aladdin.repo' "$ALADDIN_CONFIG_FILE"):$(jq -r '.aladdin.tag' "$ALADDIN_CONFIG_FILE")"}"
     local ssh_src
 
     case "$OSTYPE" in
         cygwin*) ssh_src="/.ssh" ;;
         *)       ssh_src="$(pathnorm ~/.ssh)" ;;
     esac
 
+    # How to mount docker parts
+    mounts=("${mounts[@]}" "-v /var/run/docker.sock:/var/run/docker.sock")
```
ptramsey, line 288: Same here.

aladdin.sh
```diff
@@ -323,11 +307,10 @@ function enter_docker_container() {
         -v "$(pathnorm ~/.kube):/root/.kube_local" \
         -v "$(pathnorm ~/.aladdin):/root/.aladdin" \
         -v "$(pathnorm $ALADDIN_CONFIG_DIR):/root/aladdin-config" \
-        `# Mount minikube parts` \
-        -v /var/run/docker.sock:/var/run/docker.sock \
+        ${mounts[@]} \
```
ptramsey, line 310: This should have quotes around it; else the @ is for nothing.

commands/bash/container/test-local/test-refresh.bats
```diff
@@ -2,9 +2,10 @@
 set -o pipefail
 
 @test "Test aladdin refresh command" {
-    # Call aladdin refresh 
+    # Call aladdin refresh
     $PY_MAIN refresh aladdin-test-server
     node_port=$(kubectl get svc aladdin-test-server -o json | jq -r '.spec.ports[] | select (.name == "http").nodePort')
+    local minikube_ip=$(kubectl config view -o jsonpath='{.clusters[?(@.name == "minikube")].cluster.server}' | sed -e "s/[^/]*\/\/\([^@]*@\)\?\([^:/]*\).*/\2/")
```
ptramsey, line 8: yow.  Could we use a different separator than `/`, if our pattern has so many literal forward slashes in it?  All these escapes make this pretty hard to read.  Maybe '|' or '#' ?

scripts/infra_k8s_check.sh
```diff
@@ -54,21 +54,10 @@ RESET_COLOR="$(tput sgr0 2>/dev/null || printf "\e[0m")"
 # Check things are installed
 ############################
 # Do not show the check again if it has already been installed
-ALREADY_INSTALLED_FILE="$HOME/.infra/installed"
+ALREADY_INSTALLED_FILE="$HOME/.aladdin/infra/installed"
```
ptramsey, line 57: Does it matter that we won't find the 'installed' flag from a previous version?  And does anything remain in `~/.infra`?

jarojasm95: it just means that the infra check will run once after this lands, then the `installed` flag will be reset 👍 
`~/.infra` is no longer used

commands/bash/container/test-local/test-refresh.bats
jcwilson: i've always been partial to `|` in these situations

Dockerfile
jcwilson, line 36: weird mix of tabs and spaces

commands/bash/container/test-local/test-refresh.bats
jarojasm95: cleaned this up

aladdin.sh
jarojasm95: got it, thanks

PR #48, opened 2020-05-23
Disable NFS mounts by default and optimize copying of SSH keys into minikube

jcwilson:
Cygwin users require that their `~/.ssh` directory be copied into the minikube VM to avoid
file permission issues. We only need to copy the SSH keys once, not every time aladdin is
invoked, so we now move this to the `init` procedure. The copying process has also been updated to be more resilient to repeated invocations.

Since NFS usage is an optimization and can cause issues of its own, it's now behind a default-disabled aladdin config, `nfs`. Anyone depending on it will need to `aladdin config set nfs 1` to re-enable this behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,33 +77,30 @@ Note that these commands only take effect when the minikube cluster is first set
 
 ### NFS mounts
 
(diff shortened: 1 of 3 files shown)


agianey:
Wondering if we should default to true since everyone is probably using this behavior? and then override to false? fine either way though

jcwilson:
> Wondering if we should default to true since everyone is probably using this behavior? and then override to false? fine either way though

I think having it set to not use NFS by default makes sense longer term as it reduces the friction to setting up `aladdin`. It's one less hurdle to jump through if you're just getting started.

PR #49, opened 2020-05-27
Add "componentized" build support

jcwilson:
This provides an opinionated default project image build script that allows one to organize
their project as a collection of ["components"](https://github.com/fivestars/aladdin-project-tools). This should not break any existing projects using the current method of building their docker images, as they are all already overriding this new default behavior by providing the `build_docker` value in their `lamp.json` files.

### How to use it

1. Omit the `build_docker` item from your project's `lamp.json` file.
2. Organize your project according to the guidance provided by [aladdin-project-tools](https://github.com/fivestars/aladdin-project-tools)
3. Run `aladdin build` as you normally would (or `components build`, provided by `aladdin-project-tools`)

### TODO:
- [ ] Update the `build` command documentation


lib/build-components/build_components/templates/python/macros.j2
```diff
@@ -0,0 +1,180 @@
...
+
+# Configure pip, mainly to ensure that packages are installed in the user directory.
+COPY pip.conf /etc/pip.conf
+
+# Install poetry under the root user's home directory.
+# On some images "sh" is aliased to "dash" which does not support "set -o pipefail".
+# We use the "exec" form of RUN to delegate this command to bash instead.
+# This is all because we have a pipe in this command and we wish to fail the build
+# if the any command in the pipeline fails.
+RUN ["/bin/bash", "-c", "set -o pipefail && pip install --upgrade pip && curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | POETRY_VERSION={{version}} python && chmod a+x /root/.poetry/bin/poetry"]
```
ptramsey, line 24: Why don't we want to follow this same pattern for the top-level Dockerfile?  And why is it necessary to explicitly set the poetry executable executable?  Does get-poetry.py not do that?

lib/build-components/build_components/templates/python/macros.j2
```diff
@@ -0,0 +1,180 @@
...
+{# [Builder image]
+    Install or copy the component.
+#}
+{% macro build_component(component, build_info) %}
+FROM {{build_info.builder_image}} AS builder-{{component}}
+
+# Install packages required to build native library components
+RUN apt-get update \
+ && apt-get -y --no-install-recommends install \
+    build-essential
```
ptramsey, line 39: why split across two lines?

lib/build-components/build_components/templates/python/macros.j2
```diff
@@ -0,0 +1,180 @@
...
+# We intentionally do not clear the lists here, as it's unnecessary for a builder image
+# and we may need to install some component-specific components, too.
+# The image size here isn't as much of a concern, since none of these build tools will
+# end-up in the final built image.
+
+{% if build_info.component_packages(component) %}
+# Install any additional packages required by the component
+RUN apt-get -y --no-install-recommends install \
+    {{" ".join(build_info.component_packages(component))}}
```
ptramsey, line 48: There's no other escaping we need to worry about here, right?

lib/build-components/build_components/templates/python/macros.j2
jcwilson: > Why don't we want to follow this same pattern for the top-level Dockerfile?

We probably should.

> And why is it necessary to explicitly set the poetry executable executable? Does get-poetry.py not do that?

The important thing wasn't setting it executable, it was setting it executable for everyone. Fortunately, I've figured out a way to obviate the need for this.

lib/build-components/build_components/templates/python/macros.j2
jcwilson: meh, nearly every other `apt-get ... install` we use puts all the packages on their own lines. Just looks a little weird when there's just one package.

lib/build-components/build_components/templates/python/macros.j2
jcwilson: i don't think so? did you have something in mind. The hope was that it would just be a pass-through of the `component.yaml` value. I don't think any packages have spaces in their names, which could cause issues here, i suppose.

jcwilson:
Moved to #55

Issue #50, opened 2020-06-04
[BUG] AUTHENTICATION_ENABLED: unbound variable


agianey:
If `kops validate cluster` doesn't succeed (either because the cluster is not ready, or if there is a kube-system pod that is in a failed state, we won't call `_handle_authentication_config()`, which means our `AUTHENTICATION_ENABLED` variable won't get set, which means we can't call `aladdin bash` to interact with the cluster. Instead we get:
```
/root/aladdin/commands/bash/container/bash/bash: line 12: AUTHENTICATION_ENABLED: unbound variable
```

PR #51, opened 2020-06-04
Fix AUTHENTICATION_ENABLED unbound variable bug

agianey:
To close https://github.com/fivestars-os/aladdin/issues/50

### DIFF
--- a/aladdin-container.sh
+++ b/aladdin-container.sh
@@ -91,9 +91,12 @@ function environment_init() {
 
     _handle_aws_config
 
+    # export AUTHENTICATION_ENABLED for change-permissions and bash command. By default it is false.
+    # It will only get set to true if the cluster is ready and it is enabled in aladdin-config
+    export AUTHENTICATION_ENABLED=false
+
     # Make sure we are on local or that cluster has been created before initializing helm, creating namespaces, etc
-    if "$IS_LOCAL" || ( kops export kubecfg --name $CLUSTER_NAME > /dev/null && \
-        kops validate cluster --name $CLUSTER_NAME ) > /dev/null; then
+    if _is_cluster_ready; then
 
         if "$IS_LOCAL"; then
             mkdir -p $HOME/.kube/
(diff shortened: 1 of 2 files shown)


aladdin-container.sh
```diff
@@ -91,9 +91,12 @@ function environment_init() {
...
     _handle_aws_config
 
+    # export AUTHENTICATION_ENABLED for change-permissions and bash command. By default it is false.
+    # It will only get set to true if the cluster is ready and it is enabled in aladdin-config
+    export AUTHENTICATION_ENABLED=false
+
     # Make sure we are on local or that cluster has been created before initializing helm, creating namespaces, etc
-    if "$IS_LOCAL" || ( kops export kubecfg --name $CLUSTER_NAME > /dev/null && \
-        kops validate cluster --name $CLUSTER_NAME ) > /dev/null; then
+    if _is_cluster_ready; then
```
agianey, line 99: We no longer do the kops validate cluster stuff

scripts/bash_profile.bash
```diff
@@ -25,8 +25,8 @@ function restore_permission() {
 }
 
 function with_aladdin_perms_wrapper() {
-    old_user="$(get_current_user)"
     if "$AUTHENTICATION_ENABLED"; then
+        old_user="$(get_current_user)"
```
abandali, line 29: nice cleanup here

aladdin-container.sh
```diff
@@ -122,6 +125,13 @@ function environment_init() {
 
 }
 
+function _is_cluster_ready() {
+    # Cluster is ready if we are on LOCAL or if we can pull kube config via kops
+    "$IS_LOCAL" && return 0
+    kops export kubecfg --name $CLUSTER_NAME && return 0
+    return 1
+}
```
jcwilson, line 133: IMO this reads better as

```
function _is_cluster_ready() {
    # Cluster is ready if we are on LOCAL or if we can pull kube config via kops
    "$IS_LOCAL" || kops export kubecfg --name $CLUSTER_NAME
}
```

PR #55, opened 2020-07-16
Add "componentized" build support

jcwilson:
### Note: Experimental feature
*This is an experimental feature that we're dog-fooding internally at Fivestars. We will announce
an official release once we're confident in its functionality.*

This provides an opinionated default project image build script that allows one to organize
their project as a collection of ["components"](https://github.com/fivestars/aladdin-project-tools). This should not break any existing projects using
the current method of building their docker images, as they are all already overriding this new
default behavior by providing the `build_docker` value in their `lamp.json` files.

### How to use it

1. Omit the `build_docker` item from your project's `lamp.json` file.
2. Organize your project according to the guidance provided by [aladdin-project-tools](https://github.com/fivestars/aladdin-project-tools)
3. Run `aladdin build` as you normally would (or `components build`, provided by `aladdin-project-tools`)
4. If you wish to use the standard docker build system and not buildkit, set `DOCKER_BUILDKIT=0` in your docker client's environment.


phani-kandula-fivestars:
@jcwilson  - Just to confirm my understanding. Once this PR is merged, https://github.com/fivestars/aladdin-jenkins/pull/8/files won't be necessary. We would need just one line change in Dockerfile of that repo to stop using apk add for docker. Is that correct?

And this will unblock us from the Jenkins build issues we discussed yesterday in aladdin-dev slack channel.

Dockerfile
```diff
@@ -1,43 +1,68 @@
...
-    ncurses \
-    vim
+    python3-pip
+# ncurses-bin or libncurses6 ?
 
-WORKDIR /root/aladdin
-COPY ./commands/python/requirements.txt ./commands/python/requirements.txt
-RUN pip3 install --no-cache-dir -r ./commands/python/requirements.txt
+# Default to python3 and update setuptools
+RUN ln -fs /usr/bin/python3 /usr/bin/python \
```
ptramsey, line 19: Why?

ptramsey: also might be slightly more correct to use update-alternatives if we want to do this

Dockerfile
```diff
@@ -1,43 +1,68 @@
...
+
 RUN	curl -L -o /bin/kops https://github.com/kubernetes/kops/releases/download/$KOPS_VERSION/kops-linux-amd64 \
 	&& chmod 755 /bin/kops
 
-RUN	curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && cp linux-amd64/helm \
-	/bin/helm && chmod 755 /bin/helm && helm init --client-only
+# On some images "sh" is aliased to "dash" which does not support "set -o pipefail".
+# We use the "exec" form of RUN to delegate this command to bash instead.
+# This is all because we have a pipe in this command and we wish to fail the build
+# if any command in the pipeline fails.
```
ptramsey, line 48: Blah, this is a good point. I wonder if it wouldn't be more readable just to download to a file and run the file in that case.  It might even be less code.

ptramsey: That said, even without pipefail, won't python fail if curl fails anyway?  IE, won't either part of the pipeline result in an error exit from the pipeline?

commands/python/project/project_conf.py
```diff
@@ -27,7 +27,7 @@ class ProjectConf(object):
 
     CONTENT_EXAMPLE = {
         "name": "<project_name>",
-        "build_docker": ["./commands/build.bash"],
+        # "build_docker": ["./commands/build.bash"],
```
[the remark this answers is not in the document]
ptramsey: Maybe a comment explaining why it's commented?

lib/build-components/build_components/__main__.py
```diff
@@ -0,0 +1,511 @@
...
+        logger.info(
+            "No components found for this project. Create a component directory to get started."
+        )
+        return
+
+    # Check for cycles in the component dependency graph
+    component_graph = validate_component_dependencies(components=components)
+
+    # Let's build in topological order
+    components = [
```
ptramsey, line 136: this is one case where I *might* use a new variable name, just because it's a little visually confusing to be building this list while filtering based on membership in the same list.

lib/build-components/build_components/__main__.py
```diff
@@ -0,0 +1,511 @@
...
+
+def build_aladdin_component(
+    lamp: dict,
+    build_config: BuildConfig,
+    component: str,
+    component_graph: networkx.DiGraph,
+    tag_hash: str,
+) -> None:
+    """
+    Build a component image that has been defined with a component.yaml file.
```
ptramsey, line 247: GREAT documentation

lib/build-components/build_components/__main__.py
```diff
@@ -0,0 +1,511 @@
...
+    :param cmd: The command to run.
+    :param stdin: Data to send to the subprocess as its input.
+    """
+    if stdin is None:
+        ps = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+        subprocess.run(["sed", "-e", "s/^/         /"], stdin=ps.stdout, check=True)
+        ps.wait()
+        if ps.returncode:
+            raise subprocess.CalledProcessError(ps.returncode, cmd)
+    else:
```
ptramsey, line 506: it seems like there ought to be a way to combine most of the two branches of this conditional.  Also, it ought to be possible to hook the call to sed to the end of the pipeline before calling ps.communicate() in the second case (without having to read the output into memory then separately run it through sed).   Alternatively, you could possibly skip the pipeline altogether and just use textwrap.indent().

lib/build-components/build_components/configuration.py
```diff
@@ -0,0 +1,134 @@
+import dataclasses
+import functools
+import pathlib
+import typing
+
+import yaml
+
+
+class Undefined:
```
ptramsey, line 9: This might be a good candidate to be a `@singleton`

lib/build-components/build_components/templates/python/Dockerfile.j2
```diff
@@ -0,0 +1,129 @@
+{% from "macros.j2" import install_poetry, build_component, collect_components, add_poetry, add_user, add_collected_components %}
```
ptramsey, line 1: Can't remember if I've mentioned it before, but have we looked at using an alternate build system like bazel https://github.com/bazelbuild/rules_docker instead of adding a layer of templating on top of Dockerfiles before?  Obvs out of scope here, I just stumbled across it a while back looking into how distroless worked, and it seemed neat and potentially powerful and convenient for doing complicated things building docker images.

lib/build-components/build_components/__main__.py
```diff
@@ -0,0 +1,511 @@
...
+    :param dockerfile: The dockerfile to build against. If not provided, it's assumed that a
+                       Dockerfile is present in the context directory. If it's a bytes object, it
+                       will be provided to the docker build process on stdin and a "no context"
+                       build will take place. Otherwise, a normal docker build will be performed
+                       with the specified Dockerfile.
+    """
+    buildargs = buildargs or {}
+    buildargs.setdefault("CACHE_BUST", str(time.time()))
+
+    cmd = ["env", "DOCKER_BUILDKIT=1", "docker", "build"]
```
phani-kandula-fivestars, line 469: If we enable DOCKER_BUILDKIT, it will cause failures in Jenkins build `buildkit not supported by daemon`

Dockerfile
jcwilson: > That said, even without pipefail, won't python fail if curl fails anyway? IE, won't either part of the pipeline result in an error exit from the pipeline?

Unfortunately no. You can see this in action with `false | python`. That results in a "success".

>  I wonder if it wouldn't be more readable just to download to a file and run the file in that case.

Probably, but at least in this manner we don't pollute the layer with the script file (not that it's that big in the first place).

lib/build-components/build_components/__main__.py
jcwilson: It's been a while since I've been in this code, but I do remember trying a lot of things to minimize the duplicate code here. I wasn't very successful as you can tell. I was opting for `sed` due to its ability to stream the output as it occurred, which we'd lose with `indent()`. That being said, I think `communicate()` looses that anyway, so maybe `indent()` is preferable there if we can't otherwise combine these two paths a bit more.

lib/build-components/build_components/templates/python/Dockerfile.j2
jcwilson: Yeah, I remember you mentioning it. I took a brief look at it, but it didn't seem immediately applicable to our use case. I'm not opposed to adopting it though, if it can support our needs. Perhaps it warrants a closer look.

lib/build-components/build_components/__main__.py
jcwilson: Updating our jenkins instance to version that supports it is not an option?

lib/build-components/build_components/configuration.py
jcwilson: How about if I renamed it to `_Undefined` to at least hint that it shouldn't be instantiated outside of this module and add a comment to that effect? I can think of other means of creating a throwaway class to generate a sentinel object like this, but it seems a bit overkill?

jcwilson: `del _Undefined` to the rescue

lib/build-components/build_components/__main__.py
phani-kandula-fivestars: Don't know if we can upgrade the jenkins before this can land. @jarojasm95 - your thoughts?
It seems better to not use BUILDKIT in first pass, upgrade Jenkins (in future) and bring in BUILDKIT then.
That way this PR won't get blocked on Jenkins upgrade.

jcwilson: how about if i make it configurable? defaulted to on, but you can disable?

phani-kandula-fivestars: ok. Configurable option works!

jcwilson:
```

lib/build-components/build_components/__main__.py
jcwilson: > without having to read the output into memory then separately run it through sed

Looking closer, I don't think that's what's happening. `stdout` is a file descriptor, not the contents being read completely into memory. The real sticking point is how to provide the input on `stdin`. `communicate()` was the best (only?) way I found to reasonably do that, And using that, forces us to call that prior to calling `run(['sed']) since the input there is the `stdout` from `communicate()`.

jcwilson: ...and I was mistaken. `communicate()` does in fact block and return the fully read-in stdout/stderr content. Still digging...

jcwilson: i'm going to punt on this one. Python subprocess management doesn't seem to be designed to easily replicate bash shell behavior. I'd be glad to revisit this in a later PR, though.

Dockerfile
jcwilson: I've moved these symlinks to `/usr/local/bin` so as not to interfere with any os distribution assumptions about which tools will be used.

PR #56, opened 2020-08-03
[ALDN-190] Add DNS name information to helm values for use in templates

jcwilson:
In some situations it is helpful for the templates to be able to use the domain name at which the
service will be reachable. One use case is the upcoming Ambassador `Mapping` CRDs we will be
using for edge-authenticated API calls to our services.

Additionally, we need to provide not just the namespaced service hostname, but also one for
the new concept of the "global" namespace. By default, this will be the `default` namespace but
can be overriden by providing the `service_global_namespace` value in the cluster's aladdin config.
This global hostname is useful for identifying services that are not specific to a certain namespace,
but instead exist outside of those and are common to all services in the cluster.

Finally, making the "root" hostname and arn info is useful, too.

All of these hostname values are calculateable in the same manner as the existing
`service.certificateArn` value. If the `service_dns_suffix` value is defined in the cluster's aladdin
config, it will be used for both the namespace hostname as well as the global hostname.

Ultimately, we are adding these values to the deploy-time `.Values` payload:
```
service.rootCertificateDNS
service.rootCertificateHostname
service.rootCertificateArn
service.globalCertificateDNS
service.globalCertificateHostname
service.globalCertificateArn
service.certificateDNS
service.certificateHostname
```
[ALDN-190](https://fivestars.atlassian.net/browse/ALDN-190)

### DIFF
--- a/.jira
+++ b/.jira
@@ -1,0 +1,4 @@
+[project]
+	key = ALDN
+[project "ALDN"]
+	id = 12702
--- a/.jira
+++ b/.jira
@@ -1,4 +1,0 @@
-[project]
-	key = ALDN
-[project "ALDN"]
-	id = 12702
--- a/commands/python/cluster_rules.py
+++ b/commands/python/cluster_rules.py
@@ -25,42 +25,48 @@ class ClusterRules(object):
             "'{}' object has no attribute '{}'".format(self.__class__.__name__, attr)
         )
 
-    def get_certificate_arn(self, get_global=False):
-        cert = self.values.get(
-            "service.globalCertificateArn" if get_global else "service.certificateArn"
-        )
-
-        # Check against None to allow empty string
-        if cert is None:
-            certificate_dns = self.global_certificate_dns if get_global else self.certificate_dns
-            cert = search_certificate_arn(self._boto, certificate_dns)
-        # Check against None to allow empty string
-        if cert is None:
-            cert = new_certificate_arn(self._boto, certificate_dns)
-        return cert
-
     @property
     def namespace(self):
         return self._namespace
 
     @property
-    def global_certificate_dns(self):
-        return "*.{}".format(self.global_service_dns_suffix)
+    def global_namespace(self):
+        return self.values.get("service.global_namespace", "default")
+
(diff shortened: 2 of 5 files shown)


.jira
```diff
@@ -0,0 +1,4 @@
+[project]
+	key = ALDN
+[project "ALDN"]
+	id = 12702
```
jcwilson, line 4: i'll probably back this out

commands/python/command/deploy.py
```diff
@@ -101,6 +101,11 @@ def deploy(
         # Values precedence is command < cluster rules < --set-override-values
         # Deploy command values
         values = {
+            "service.globalCertificateDNS": cr.global_certificate_dns,
+            "service.globalCertificateHostname": cr.global_certificate_dns.strip("*."),
+            "service.certificateDNS": cr.certificate_dns,
+            "service.certificateHostname": cr.certificate_dns.strip("*."),
+            "service.globalCertificateArn": cr.get_certificate_arn(get_global=True),
```
jcwilson, line 108: could really use some suggestions for better names here

jarojasm95: I'm not a big fan of using "global" prefix for variable identifiers, maybe call it an override instead?

commands/python/cluster_rules.py
```diff
@@ -25,31 +25,62 @@ def __getattr__(self, attr):
...
         """
         return "{}.{}".format(self._namespace, self.root_dns)
 
+    @property
+    def global_service_dns_suffix(self):
+        """
+        The DNS name for the cluster-wide global namespace.
+
+        To be used for "cluster global" resources when deploying to a cluster that makes use of
+        namespaces to separate deployments.
```
jarojasm95, line 78: maybe "primary" namespace?

jcwilson:
Closing this to rethink some concepts here. I'll post a new PR once that settles.

commands/python/command/deploy.py
jcwilson: It's not really an override, though.
It might be a moot point, as this global namespace is looking more like an anti-pattern that we may not want to encourage.

PR #57, opened 2020-08-05
[ALDN-192] Fix chown of WORKDIR to USER when publishing

jcwilson:
We need to do the `chown` before switching the `USER`.

[ALDN-192](https://fivestars.atlassian.net/browse/ALDN-192)


lib/build-components/build_components/templates/python/macros.j2
```diff
@@ -170,9 +170,6 @@ RUN groupadd -f {{user_info.group}} \
  && rm -rf /var/lib/apt/lists/* \
  && echo >> /etc/sudoers "{{user_info.name}} ALL=(ALL) NOPASSWD: ALL"
 {% endif %}
-
```
phani-kandula-fivestars, line 173: So I see only removing the USER command. Do we add it somewhere later in the file?

phani-kandula-fivestars: oops. it's in the other file. I'll read more closely.

PR #59, opened 2020-08-06
[ALDN-193] Support deploying a specific chart in a project

jcwilson:
In some situations, a project may need to maintain two or more charts. `aladdin start`, `aladdin publish`
and `aladdin deploy` all currently support an all-or-nothing approach to their respective operations.
We should be able to specify which chart we wish to deploy.

This adds the `--chart` option to the `deploy', 'restart', 'rollback', 'start', 'stop', and 'undeploy' commands.
The default behavior for `start`, `stop` and `restart` is to perform the operation on all charts if not
otherwise specified. For the others, it will default to using the project's name (which prior to this was a
hard-coded assumption) and will only ever use a single chart per invocation. Multiple invocations are
required to deploy multiple charts.

This also adds the `.Values.project.name` and `.Values.deploy.namespace` values to the helm templating
process during the `start`, `restart` and `deploy` commands.

`aladdin publish` will now push all charts individually to S3 (there is no need to specify the chart to publish)
and `aladdin deploy` now pulls all charts down before deciding which one to publish according to the
`--chart` value.

#### TODO:
- [x] Add `--chart` option to remaining related commands, too
- [x] Update docs

[ALDN-193](https://fivestars.atlassian.net/browse/ALDN-193)

### DIFF
--- a/commands/python/command/restart.py
+++ b/commands/python/command/restart.py
@@ -11,7 +11,7 @@ def parse_args(sub_parser):
         "--chart",
         action="append",
         dest="charts",
-        help="Only start these charts (may be specified multiple times)",
+        help="Start only these charts (may be specified multiple times)",
     )
     subparser.add_argument(
         "--with-mount",
@@ -26,5 +26,5 @@ def restart_args(args):
 
 
 def restart(namespace, charts, with_mount):
-    stop.stop(namespace)
+    stop.stop(namespace, charts)
     start.start(namespace, charts, False, with_mount)
--- a/commands/python/command/restart.py
+++ b/commands/python/command/restart.py
@@ -8,6 +8,12 @@ def parse_args(sub_parser):
     subparser.set_defaults(func=restart_args)
     add_namespace_argument(subparser)
     subparser.add_argument(
+        "--chart",
+        action="append",
+        dest="charts",
+        help="Only start these charts (may be specified multiple times)",
+    )
+    subparser.add_argument(
         "--with-mount",
         "-m",
         action="store_true",
@@ -16,9 +22,9 @@ def parse_args(sub_parser):
 
 
 def restart_args(args):
-    restart(args.namespace, args.with_mount)
+    restart(args.namespace, args.charts, args.with_mount)
 
 
-def restart(namespace, with_mount):
+def restart(namespace, charts, with_mount):
     stop.stop(namespace)
-    start.start(namespace, False, with_mount)
+    start.start(namespace, charts, False, with_mount)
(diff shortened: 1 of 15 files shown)


commands/python/command/restart.py
```diff
@@ -16,9 +22,9 @@ def parse_args(sub_parser):
...
 def restart_args(args):
-    restart(args.namespace, args.with_mount)
+    restart(args.namespace, args.charts, args.with_mount)
 
 
-def restart(namespace, with_mount):
+def restart(namespace, charts, with_mount):
     stop.stop(namespace)
-    start.start(namespace, False, with_mount)
+    start.start(namespace, charts, False, with_mount)
```
phani-kandula-fivestars, line 30: stop everything in the namespace or just the charts (if not None)? 
Asking because help says 'Only start these charts..' does it imply we stop only these charts

commands/python/command/start.py
```diff
@@ -62,9 +88,11 @@ def start(namespace, dry_run=False, with_mount=False, force_helm=False, set_over
...
-        sync_ingress.sync_ingress(namespace)
-        sync_dns.sync_dns(namespace)
+    for chart_path in pc.get_helm_chart_paths():
+        if os.path.basename(chart_path) in charts:
+            if dry_run:
+                helm.dry_run(hr, chart_path, cr.cluster_name, namespace, **values)
+            else:
+                helm.start(hr, chart_path, cr.cluster_name, namespace, force_helm, **values)
+                sync_ingress.sync_ingress(namespace)
+                sync_dns.sync_dns(namespace)
```
phani-kandula-fivestars, line 98: can sync ingress and sync dns be done just once after all charts have been started? Asking because of for loop at line 91

commands/python/libs/k8s/helm.py
```diff
@@ -3,15 +3,16 @@
...
 from botocore.exceptions import ClientError
-from os.path import join, dirname, expanduser
+from os.path import join, expanduser
 
 logger = logging.getLogger(__name__)
 
 
 class Helm(object):
-    PACKAGE_PATH = "helm_charts/0.0.0/{project_name}/{git_ref}/{project_name}.{git_ref}.tgz"
-    VALUE_PATH = "helm_charts/0.0.0/{project_name}/{git_ref}/values.{values_name}.yaml"
```
ptramsey, line 14: Did nothing use this?

jcwilson: not that I could find

commands/python/command/restart.py
jcwilson: good call. let me take a pass at that

commands/python/command/start.py
jcwilson: yeah, i think i'll move it out and only perform the sync if any `helm.start()` call succeeds.

PR #60, opened 2020-08-07
[ALDN-190] Add DNS and TLS cert information to helm template values

jcwilson:
In some situations it is helpful for the templates to be able to use the domain name the service
will be reachable at. One use case is the upcoming Ambassador `Mapping` CRDs we will be using for
edge-authenticated API calls to our services.

Additionally, it is useful in some situations to also have the same information for the
non-namespaced cluster DNS name.

This adds some values alongside the existing `.Values.service.certificateArn` value:

| .Values name | Notes |
| --- | --- |
| `service.clusterDomainName` | The top-level, non-namespaced, non-service-specific DNS hostname for the cluster |
| `service.clusterCertificateScope` | The wildcard TLS certificate scope for the above |
| `service.clusterCertificateArn` | The AWS ARN of the certificate for that scope |
| `service.certificateHostname` | The namespaced DNS hostname for the cluster's services |
| `service.certificateScope` | The wildcard TLS certificate scope for the above |
| `service.domainName` | The potentially namespaced domain name under which the service hostnames would be found |

[ALDN-190](https://fivestars.atlassian.net/browse/ALDN-190)


commands/python/cluster_rules.py
```diff
@@ -86,23 +109,6 @@ def dual_dns_prefix_annotation_name(self):
     def _boto(self):
         return boto3.Session(profile_name=self.aws_profile)
 
-    def fill_hostedzone(self, services_by_name):
```
jcwilson, line 89: I moved this to `lib.aws.dns_mapping`.

commands/python/cluster_rules.py
```diff
@@ -25,34 +18,64 @@ def __getattr__(self, attr):
             "'{}' object has no attribute '{}'".format(self.__class__.__name__, attr)
         )
 
-    def get_certificate_arn(self):
+    def get_certificate_arn(self, for_cluster=False):
```
jcwilson, line 21: I could probably be persuaded to break this into two functions that call a common internal function to keep the interface a bit more explicit

commands/python/command/deploy.py
```diff
@@ -101,6 +101,11 @@ def deploy(
         # Values precedence is command < cluster rules < --set-override-values
         # Deploy command values
         values = {
+            "service.clusterCertificateHostname": cr.cluster_domain_name_suffix,
```
jcwilson, line 104: it's not really a "certificate" hostname. I'm thinking of changing these to just `service.clusterHostname`  and `service.hostname`.

commands/python/command/sync_dns.py
```diff
@@ -23,11 +24,19 @@ def sync_dns(namespace):
...
+
+    service_loadbalancers = k_utils.get_services_to_load_balancers_map(
+        cr.dual_dns_prefix_annotation_name, cr.ingress_info
     )
+
+    # Apply our dns to the service names
+    service_hostnames_to_loadbalancers = {
+        f"{service_name}.{cr.namespace_domain_name}": loadbalancer_hostname
+        for service_name, loadbalancer_hostname in service_loadbalancers.items()
+    }
```
jcwilson, line 38: Brought this in from `ClusterRules.fill_hostedzone()` when it got moved to `dns_mapping`

Issue #64, opened 2020-08-28
Allow option to skip installing software


phani-kandula-fivestars:
Aladdin installs minikube, jq etc

But if I have another solution in place to manage configuration of my machine (like ansible playbook for example) I would continue to use that and still use aladdin for rest of the use cases.

Since current behavior is to install software. We can make it the default option.

Something like INSTALL_SOFTWARE=true and users can override with INSTALL_SOFTWARE=false when they invoke aladdin if they choose to.

PR #65, opened 2020-08-28
[ISSUE-64] Allow option to skip installing software

phani-kandula-fivestars:
Bare minimum changes to make software installation optional.
Not trying to refactor the bash script at this time.

[ISSUE-64](https://github.com/fivestars-os/aladdin/issues/64)

### DIFF
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -15,6 +15,7 @@ NAMESPACE=default
 IS_TERMINAL=true
 SKIP_PROMPTS=false
 KUBERNETES_VERSION="1.15.6"
+INSTALL_SOFTWARE=true
 
 # Set key directory paths
 ALADDIN_DIR="$(cd "$(dirname "$0")" ; pwd)"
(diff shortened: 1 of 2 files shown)


aladdin.sh
```diff
@@ -373,6 +378,9 @@ while [[ $# -gt 0 ]]; do
         --skip-prompts)
             SKIP_PROMPTS=true
         ;;
+        --no-install-software)
```
phani-kandula-fivestars, line 381: @jcwilson - I can get this value also from aladdin config like I do for MANAGE_MINIKUBE is a separate PR.
Will push that commit tomorrow.

Reason: Choosing to not let aladdin do config management seems more permanent decision .. similar to choosing MANAGE_MINIKUBE = false

aladdin.sh
```diff
@@ -52,6 +53,14 @@ function get_plugin_dir() {
     fi
 }
 
+function get_install_software() {
```
phani-kandula-fivestars, line 56: `aladdin config set install_software false` once is all we need to skip software installation for all future invocations of aladdin.

jarojasm95 (approved):

ptramsey (approved):

Issue #66, opened 2020-08-28
Allow option to ask aladdin to not manage my minikube instance


phani-kandula-fivestars:
Aladdin sets up minikube in a certain way:

1. Set up nfs mounting, 
2. Set config based on config set in aladdin with minikube.key style etc.

If I already have a way to manage minikube, I would prefer to do it using my existing setup and prefer to use aladdin for rest of the use cases.

Perhaps an argument like --no-manage-minikube? That way, if that flag isn't passed, aladdin can proceed with the current/default behavior.

PR #67, opened 2020-08-28
[ISSUE-66] Option to ask aladdin to not manage minikube

phani-kandula-fivestars:
minikube commands are executed with direct calls to `minikube`
and with helper function.

When --no-manage-minikube is set:
1. direct calls to `minikube` become no op.
2. helper function return right away if MANAGE_MINIKUBE is false

This change might seem hacky. But  aladdin.sh will have to be
rewritten first to make it easy to make these types of changes.

[ISSUE-66](https://github.com/fivestars-os/aladdin/issues/66)

### DIFF
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -14,6 +14,7 @@ CLUSTER_CODE=minikube
 NAMESPACE=default
 IS_TERMINAL=true
 SKIP_PROMPTS=false
+MANAGE_MINIKUBE=true
 KUBERNETES_VERSION="1.15.6"
 
 # Set key directory paths
@@ -30,6 +31,15 @@ DEFAULT_MINIKUBE_MEMORY=4096
 
 source "$SCRIPT_DIR/shared.sh" # to load _extract_cluster_config_value
 
+# If $MANAGE_MINIKUBE is true, then all commands will be executed
+# using the external 'minikube' command.
+# https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#index-command
+# Otherwise calls to minikube will be no-op
+function minikube() {
+    if "$MANAGE_MINIKUBE"; then
+      command minikube "$@"
+    fi
+}
 function get_config_path() {
     if [[ ! -f "$HOME/.aladdin/config/config.json" ]]; then
         echo "Unable to find config directory. Please use 'aladdin config set config_dir <config path location>' to set config directory"
@@ -92,6 +102,9 @@ function check_and_handle_init() {
 }
 
 function set_minikube_config(){
+    if [ "$MANAGE_MINIKUBE" = false ] ; then
+        return
+    fi
     minikube config set kubernetes-version v$KUBERNETES_VERSION > /dev/null
 
     for key in vm_driver memory disk_size cpus; do
@@ -107,6 +120,9 @@ function set_minikube_config(){
 }
 
 function _start_minikube() {
+    if [ "$MANAGE_MINIKUBE" = false ] ; then
+        return
+    fi
     local minikube_cmd="minikube start"
 
     if test "$OSTYPE" = "linux-gnu"; then
@@ -132,6 +148,9 @@ function _start_minikube() {
 
 # Start minikube if we need to
 function check_or_start_minikube() {
+    if [ "$MANAGE_MINIKUBE" = false ] ; then
+        return
+    fi
     if ! minikube status | grep Running > /dev/null; then
 
         echo "Starting minikube... (this will take a moment)"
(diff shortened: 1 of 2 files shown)


aladdin.sh
```diff
@@ -30,6 +31,15 @@ DEFAULT_MINIKUBE_MEMORY=4096
 
 source "$SCRIPT_DIR/shared.sh" # to load _extract_cluster_config_value
 
+# If $MANAGE_MINIKUBE is true, then all commands will be executed
+# using the external 'minikube' command.
+# https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#index-command
+# Otherwise calls to minikube will be no-op
+function minikube() {
+    if "$MANAGE_MINIKUBE"; then
```
jcwilson, line 39: let's use this conditional form throughout, rather than `if [ "$MANAGE_MINIKUBE" = false ] ; then`

ptramsey (commented):
With the exception of josh's comment w.r.t. aladdin config, this looks fine to me.

aladdin.sh
phani-kandula-fivestars: Pushed the commit with this suggestion implemented.

aladdin.sh
```diff
@@ -52,6 +62,15 @@ function get_plugin_dir() {
     fi
 }
 
+function get_manage_minikube() {
```
phani-kandula-fivestars, line 65: now getting MANAGE_MINIKUBE value from config.
Must be set like this: `aladdin config set manage_minikube false`
`true` is the default value because that's how aladdin works today.

aladdin.sh
```diff
@@ -30,6 +31,15 @@ DEFAULT_MINIKUBE_MEMORY=4096
 
 source "$SCRIPT_DIR/shared.sh" # to load _extract_cluster_config_value
 
+# If $MANAGE_MINIKUBE is true, then all commands will be executed
+# using the external 'minikube' command.
+# https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#index-command
+# Otherwise calls to minikube will be no-op
+function minikube() {
```
jarojasm95, line 38: I am a bit confused, we're adding this option but then we're also doing `if "$MANAGE_MINIKUBE";` too, do we need to do both?

jarojasm95 (commented):
I think we might be able to better serve this purpose by just using `CLUSTER_CODE`, meaning if `CLUSTER_CODE` is `minikube` then run all the minikube commands (make sure it's started, configured, etc), and otherwise those can be skipped. We could also add a config value for what the user wants to define as the "local cluster" or "default cluster" which would let the user use any cluster setup as long as the kube context can be found 😉 

I am okay with current path this is taking, just thought of this different approach too^. Feel free to disregard if you have a strong opinion.

ptramsey:
Hmmm... it's a good point.  Because in this world, what we're moving towards is a world where 'minikube' is shorthand for "local development", which may or may not actually involve minikube.

phani-kandula-fivestars:
> I think we might be able to better serve this purpose by just using `CLUSTER_CODE`, meaning if `CLUSTER_CODE` is `minikube` then run all the minikube commands (make sure it's started, configured, etc), and otherwise those can be skipped. We could also add a config value for what the user wants to define as the "local cluster" or "default cluster" which would let the user use any cluster setup as long as the kube context can be found 😉
> 
> I am okay with current path this is taking, just thought of this different approach too^. Feel free to disregard if you have a strong opinion.

CLUSTER_CODE=minikube and manage.minikube=true must always go together. Otherwise user will have to setup nfs etc.
However skipping minikube completely if CLUSTER_CODE!=minikube (say KDEV) might be tricky. If user has no docker installed, they still need minikube to provide docker.

In future we can further split what manage.minikube means to clarify if we want local docker or thru minikube etc. if we want nfs mounting etc. Just need more config vars. Some will go under minikube section of config and some under manage. 

Happy to discuss. We use minikube for multiple reasons. So this discussion is natural and good!

ptramsey (approved):

jarojasm95 (approved):

jarojasm95:
I'm creating a separate issue to address some of the follow up items here:
- renaming the default cluster to something other than "minikube" since that may not always be how local development is setup
- make that configurable^
- incorporate all this into aladdin docs^, and maybe add a section about how to use other setups? (docker-desktop, etc)

phani-kandula-fivestars:
Created this Issue https://github.com/fivestars-os/aladdin/issues/68 -- to follow up on some of the conversations in this PR.

PR #73, opened 2020-09-21
Make aladdin an installable python package

jarojasm95:
This PR does a lot of reorganization mostly in order for aladdin to fit into what poetry expects of a python package. It is intended to make porting over bash scripts to python smoother, and transitions aladdin to be a python package with a cli entrypoint.

- install script
This install script uses [pipx](https://github.com/pipxproject/pipx), a python library used to install cli tools while wrapping them in their own venv. This allows aladdin to have strict and very well defined python dependencies and not have to worry about them clashing with the user's system packages.

- merge of `build-components` into aladdin commands
This was done because pipx was having issues figuring out our project-in-project setup, I don't think this should affect any functionality.

Additionally there's some opportunistic improvements to the Dockerfile.

How to try it out:
- `git checkout pipx`
- `$(minikube docker-env)`
- `docker build -t aladdin:dev .`
- update your aladdin config repo/tag to use `aladdin:dev`
- `install-aladdin`

TODO:
- update documentation
- something else??


Dockerfile
```diff
@@ -7,23 +31,18 @@ RUN apt-get update && \
...
-RUN ln -fs /usr/bin/python3 /usr/local/bin/python && \
-    ln -fs /usr/bin/pip3 /usr/local/bin/pip && \
-    pip install --no-cache-dir setuptools==46.4.0 wheel==0.34.2
+RUN pip install --no-cache-dir pip==20.2.3
 
-# This can take a bit of time, so we do it earlier in the build process
-RUN go get -u -v sigs.k8s.io/aws-iam-authenticator/cmd/aws-iam-authenticator
+ARG AWS_IAM_AUTHENTICATOR_VERSION=1.17.9
+RUN curl -o /usr/local/bin/aws-iam-authenticator https://amazon-eks.s3.us-west-2.amazonaws.com/$AWS_IAM_AUTHENTICATOR_VERSION/2020-08-04/bin/linux/amd64/aws-iam-authenticator && \
+    chmod 755 /usr/local/bin/aws-iam-authenticator
```
jarojasm95, line 45: doing `go get` was taking a LOT of time heh

aladdin.sh
```diff
@@ -113,7 +113,9 @@ function check_and_handle_init() {
             eval "$repo_login_command"
         fi
         local aladdin_image="$(jq -r '.aladdin.repo' "$ALADDIN_CONFIG_FILE"):$(jq -r '.aladdin.tag' "$ALADDIN_CONFIG_FILE")"
-        docker pull "$aladdin_image"
+        if [[ $aladdin_image == *"/"* ]]; then
+            docker pull "$aladdin_image"
```
jarojasm95, line 117: only do `docker pull` if the image is in the form: `repo/image:tag`, if there's no repo then don't do `docker pull`, maybe there's a better way to do this?

aladdin/commands/build_components/__main__.py
```diff
@@ -74,7 +41,7 @@ def main():
         lamp=lamp,
         tag_hash=os.getenv("HASH", "local"),
         build_config=BuildConfig(),
-        components=sys.argv[1:],
+        components=sys.argv[2:],
```
jarojasm95, line 44: the command is not `aladdin build-components ...` so skip the first two args

aladdin/arg_tools.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+
+def bash_wrapper():
+    _, *args = sys.argv
+    handler = os.path.join(PROJECT_ROOT, "aladdin.sh")
+    subprocess.run([handler, *args])
+
+
+def container_command(func=None):
```
jarojasm95, line 27: one of the most significant additions in this PR, it redirects a command over to the bash scripts, then the same command is invoked inside the container where it's actually executed

aladdin/lib/logging.py
```diff
@@ -0,0 +1,7 @@
+import logging
+
+import lazy_object_proxy
+
+
+def getLogger(*args, **kwargs):
+    return lazy_object_proxy.Proxy(lambda: logging.getLogger(*args, **kwargs))
```
jarojasm95, line 7: used so that any `logger` instances created at the module level are lazily evaluated and get an instance of `VerboseLogs`

Dockerfile
```diff
@@ -53,31 +69,21 @@ RUN curl -L -o /usr/local/bin/kops https://github.com/kubernetes/kops/releases/d
...
 
 WORKDIR /root/aladdin
 
-# Install aladdin python requirements
-COPY pyproject.toml poetry.lock ./
-COPY lib/build-components /root/aladdin/lib/build-components
-RUN poetry install --no-root
-
+COPY --from=code /root/.poetry /root/.poetry
+COPY --from=code /root/.venv /root/.venv
```
jcwilson, line 84: why are we bothering with `.venv`?

aladdin/commands/build_components/__main__.py
jcwilson: this is an interface breaking change, correct?

poetry.toml
```diff
@@ -1,7 +1,3 @@
-# This file is copied into our docker images to globally configure the poetry tool
-
 [virtualenvs]
-# We're in a docker container, there's no need for virtualenvs
-# One should still specify "ENV PIP_USER yes" to let poetry know
-# install packages as --user so they show up in ~/.local
-create = false
+create = true
+in-project = true
```
jcwilson, line 3: are we sure this is what we want? Is this evaluated on the host machine? or in the Dockerfile build?

jcwilson (commented):
Whether a module gets placed in `aladdin.lib` or `aladdin` seems rather arbitrary. Can we clean that up, either in this PR or a fast-follow before these changes are released?

Dockerfile
jarojasm95: it makes docker multistage easier, since we know everything is in that dir

aladdin/commands/build_components/__main__.py
jarojasm95: hmm not quite, I think, outside the container the command is still `aladdin build-components` and inside the container `build-components` also still works

poetry.toml
jarojasm95: this is strictly used for aladdin development, it could be removed all together and leave it up to the devs global config, I don't have a strong opinion so I might just remove it

aladdin/commands/build_components/__main__.py
jcwilson: as long as `aladdin build` still runs this code when `build_docker` is not specified in the `lamp.json` we should be ok

jarojasm95: yep still works 👍 I tried it on reporting-service

jarojasm95:
I can move everything into `lib` as a fast-follow after this PR

Dockerfile
```diff
@@ -1,4 +1,24 @@
-FROM golang:1.14-buster
+FROM python:3.8.5-buster as code
```
ptramsey, line 1: `code` seems a little ambiguous to me, since the final image contains our code, too.  Would maybe `build` make sense?

Dockerfile
```diff
@@ -1,4 +1,24 @@
...
+# This is all because we have a pipe in this command and we wish to fail the build
+# if any command in the pipeline fails.
+ARG POETRY_VERSION=1.0.10
+RUN ["/bin/bash", "-c", "set -o pipefail && curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python"]
+ENV PATH /root/.poetry/bin:$PATH
+
+ARG POETRY_VIRTUALENVS_CREATE="false"
+# Install aladdin python requirements
+COPY pyproject.toml poetry.lock ./
+RUN poetry install --no-root
```
ptramsey, line 19: What is gained by running just this in a separate image?  Are we avoiding garbage left in the filesystem somewhere by installing here and then copying over .venv?

aladdin.sh
```diff
@@ -312,24 +314,22 @@ function handle_ostypes() {
...
-        fi
+        MOUNTS_CMD="-v $(pathnorm "$ALADDIN_DIR"):/root/aladdin"
     fi
 
-    # If on minikube mount minikube credentials
     if "$IS_LOCAL"; then
-        MINIKUBE_CMD=""
         if [ "$CLUSTER_CODE" = "minikube" ] ; then
-            MINIKUBE_CMD="-v $(pathnorm ~/.minikube):/root/.minikube"
-            MINIKUBE_CMD="$MINIKUBE_CMD -v /:/minikube_root" # mount the whole minikube system
```
ptramsey, line 328: ick

aladdin/__init__.py
```diff
@@ -0,0 +1,2 @@
+VERSION = (1, 15, 6, 12)
```
ptramsey, line 1: will we have something to automatically update this?

aladdin/commands/build_components/__init__.py
```diff
@@ -0,0 +1,8 @@
+from aladdin.commands.build_components import __main__
```
ptramsey, line 1: If build_components is an aladdin command now... presumably, it won't be directly executed as a python program anymore.  Could __main__.py be renamed, since that has specific implications?  Or is there a reason that it shouldn't be?

aladdin/commands/build_components/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+
+
+def parse_args(sub_parser):
+    subparser = sub_parser.add_parser(
+        "build-components", help=__main__.__doc__
+    )
+    subparser.set_defaults(func=lambda args: __main__.main())
```
ptramsey, line 8: is there a reason we don't pass args to main()?  This seems slightly counterintuitive.

ptramsey (commented):
Otherwise looks good!

aladdin/commands/build_components/__init__.py
jarojasm95: it can be renamed 👍

aladdin/__init__.py
jarojasm95: suggestions welcome!

Dockerfile
jarojasm95: the total aladdin image gets to be way smaller! we need to use debian-buster for installation of requirements but then drop down to debian-buster-slim for the final image

Dockerfile
jarojasm95: sure 👍

ptramsey (approved):

PR #74, opened 2020-10-27
Use helm3 in aladdin commands with fallback helm2 option

agianey:
During the helm3 migration, we will have aladdin commands use helm3, but have support for helm2. Once everything has been confirmed to migrate over, we will remove tiller, and remove helm2 support from aladdin.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,16 +36,16 @@ ARG KUBE_VERSION=1.15.6
 RUN curl -L -o /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v$KUBE_VERSION/bin/linux/amd64/kubectl && \
     chmod 755 /usr/local/bin/kubectl
 
-ARG HELM_VERSION=2.16.1
-RUN curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && \
-    cp linux-amd64/helm /usr/local/bin/helm && \
-    chmod 755 /usr/local/bin/helm && \
-    helm init --client-only
+ARG HELM2_VERSION=2.16.1
+RUN curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM2_VERSION-linux-amd64.tar.gz | tar -zxvf - && \
+    cp linux-amd64/helm /usr/local/bin/helm2 && \
+    chmod 755 /usr/local/bin/helm2 && \
+    helm2 init --client-only
 
-ARG HELM3_VERSION=3.3.0
-RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 && \
-    chmod 700 get_helm.sh && \
-    BINARY_NAME=helm3 ./get_helm.sh --version v$HELM3_VERSION
+ARG HELM_VERSION=3.3.4
+RUN curl -L -o- https://get.helm.sh/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && \
+    cp linux-amd64/helm /usr/local/bin/helm && \
+    chmod 755 /usr/local/bin/helm
 
 ARG KOPS_VERSION=1.15.0
 RUN curl -L -o /usr/local/bin/kops https://github.com/kubernetes/kops/releases/download/$KOPS_VERSION/kops-linux-amd64 && \
--- a/aladdin-container.sh
+++ b/aladdin-container.sh
@@ -95,7 +95,7 @@ function environment_init() {
     # It will only get set to true if the cluster is ready and it is enabled in aladdin-config
     export AUTHENTICATION_ENABLED=false
 
-    # Make sure we are on local or that cluster has been created before initializing helm, creating namespaces, etc
+    # Make sure we are on local or that cluster has been created before creating namespaces, etc
     if _is_cluster_ready; then
 
         if "$IS_LOCAL"; then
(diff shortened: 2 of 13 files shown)


Dockerfile
```diff
@@ -36,16 +36,16 @@ ARG KUBE_VERSION=1.15.6
 RUN curl -L -o /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v$KUBE_VERSION/bin/linux/amd64/kubectl && \
     chmod 755 /usr/local/bin/kubectl
 
-ARG HELM_VERSION=2.16.1
-RUN curl -L -o- https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && \
```
agianey, line 40: have helm2 be helm2 rather than helm

Dockerfile
```diff
@@ -36,16 +36,16 @@ ARG KUBE_VERSION=1.15.6
...
+    helm2 init --client-only
+
+ARG HELM_VERSION=3.3.4
+RUN curl -L -o- https://get.helm.sh/helm-v$HELM_VERSION-linux-amd64.tar.gz | tar -zxvf - && \
     cp linux-amd64/helm /usr/local/bin/helm && \
-    chmod 755 /usr/local/bin/helm && \
-    helm init --client-only
-
-ARG HELM3_VERSION=3.3.0
-RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 && \
```
agianey, line 46: helm now uses helm3

aladdin-container.sh
```diff
@@ -115,7 +115,6 @@ function environment_init() {
 
         if $INIT; then
             kubectl create namespace --cluster $CLUSTER_NAME $NAMESPACE || true
-            _initialize_helm
```
agianey, line 118: server side of helm is gone in helm3, so we don't need to do this

commands/bash/container/delete-namespace/delete-namespace
```diff
@@ -2,7 +2,7 @@
 set -eu -o pipefail
 
 function delete_namespace {
-    helm delete --purge $(helm list | awk '{print $1}' | grep -E "\-$1\b")
```
agianey, line 5: --purge option is gone in helm3

commands/bash/container/delete-namespace/delete-namespace
```diff
@@ -2,7 +2,7 @@
...
+    helm delete -n $1 $(helm list --short -n $1)
```
agianey, line 5: helm is namespace scoped now

commands/python/command/cluster_init.py
```diff
@@ -18,15 +18,20 @@ def parse_args(sub_parser):
         action="store_true",
         help="don't check if cluster init projects are already installed before installing",
     )
+    subparser.add_argument(
```
agianey, line 21: Support helm2 options on aladdin commands during migration in case something fails to move over

commands/python/libs/k8s/helm.py
```diff
@@ -131,23 +155,36 @@ def release_exists(self, release_name):
...
         # Some ugly logic to get the current revision - probably can be better
-        output = subprocess.Popen(("helm", "list"), stdout=subprocess.PIPE)
+        helm_list_command = [self.helm, "list"]
+        # Helm3 requires --namespace flag, which helm2 doesn't use
+        if self.helm == "helm":
+            helm_list_command.extend(["--namespace", namespace])
+        output = subprocess.Popen(helm_list_command, stdout=subprocess.PIPE)
         output = subprocess.check_output(("grep", release_name), stdin=output.stdout)
-        current_revision = int(output.decode("utf-8").replace(" ", "").split("\t")[1])
+        data = output.decode("utf-8").replace(" ", "").split("\t")
```
edgar-ochoa-fivestars, line 168: nit: this seems to be working fine but would it be better to add a "--output json|yaml" flag on the command to explicitly grab the revision versus selecting a column below?

agianey: Good suggestion - I will take that on in the forthcoming PR which will remove helm2 completely if that works

edgar-ochoa-fivestars (approved):
LGTM!

jarojasm95 (approved):
🥳

commands/bash/container/delete-namespace/delete-namespace
jcwilson: will we still be appending the namespaces to the helm releases moving forward?

aladdin-container.sh
jcwilson: If one does need to use helm2 for some reason, they would need to do this manually, correct? But it seems we're removing this helper function entirely.

commands/python/libs/k8s/helm.py
```diff
@@ -14,13 +14,19 @@ class Helm(object):
...
     def helm_home(self):
         return join(expanduser("~"), ".helm")
 
     def init(self):
-        # We need to have local helm initialized for it to works
-        subprocess.check_call(["helm", "init", "--client-only"])
+        # Only need to init on helm2. Helm2 Doesn't use tiller and therefore needs no init
+        if self.helm == "helm2":
+            # We need to have local helm2 initialized for helm2 to work
+            subprocess.check_call([self.helm, "init", "--client-only"])
```
jcwilson, line 29: I'm confused how this works if we fully removed the tiller setup in `aladdin.sh`, but i'm not saying it won't.

jcwilson (approved):

PR #78, opened 2020-11-07
Add support for cluster config values

jarojasm95:
This PR adds support for a `values.yaml` file in `aladdin-config`, it provides the same functionality as the `"values"` key in `config.json`, however a yaml file provides more flexibility in the data types that are supported and how the data is structured.

### DIFF
--- a/aladdin/lib/k8s/helm.py
+++ b/aladdin/lib/k8s/helm.py
@@ -115,6 +115,15 @@ class Helm(object):
         if os.path.isfile(cluster_config_values_path):
             logger.info("Found cluster config values file")
             values.append(cluster_config_values_path)
+        cluster_namespace_config_values_path = os.path.join(
+            os.environ["ALADDIN_CONFIG_DIR"],
+            cluster_name,
+            namespace,
+            "values.yaml"
+        )
+        if os.path.isfile(cluster_namespace_config_values_path):
+            logger.info("Found cluster namespace config values file")
+            values.append(cluster_namespace_config_values_path)
 
         return values
 
--- a/aladdin/lib/k8s/helm.py
+++ b/aladdin/lib/k8s/helm.py
@@ -204,11 +204,13 @@ class Helm(object):
         # Values and overrides are defined/specified in helm args in a specific order
         # 1. project values.yaml (picked up by helm automatically)
         # 2. project cluster values.yaml
-        # 3. site.yaml file (on local)
-        # 4. aladdin-config cluster `values.yaml`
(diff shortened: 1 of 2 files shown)


jcwilson:
Would it be possible to replace (or augment) the current behavior that loads `config.json` files from the aladdin config directory to also look for yaml files? that way we can keep the existing structure, just in yaml, rather than json. First one found wins, and maybe preference is given to yaml?

The proposal here means we now have two places to define cluster[+namespace] values and I could see that causing confusion.

aladdin/lib/k8s/helm.py
```diff
@@ -107,6 +107,24 @@ def find_values(self, chart_path, cluster_name, namespace):
...
+        cluster_namespace_config_values_path = os.path.join(
+            os.environ["ALADDIN_CONFIG_DIR"],
+            cluster_name,
+            namespace,
+            "values.yaml"
+        )
+        if os.path.isfile(cluster_namespace_config_values_path):
+            logger.info("Found cluster namespace config values file")
+            values.append(cluster_namespace_config_values_path)
+
```
jcwilson, line 127: I think order here is important, so maybe these should be added to the values list before any project configuration?

jcwilson: let's also define a local consistency here where we define the file at time of use for each `if` block, or we define them all at once at the top

jarojasm95: I've taken order into account 👍 I've added a comment to make it explicit/clear. I also moved the variable definition close to the `if` block.

aladdin/lib/k8s/helm.py
```diff
@@ -183,6 +201,15 @@ def _run(self, helm_rules, chart_path, cluster_name, namespace, helm_args=None,
             "--namespace={}".format(namespace),
         ]
 
+        # Values and overrides are defined/specified in helm args in a specific order
+        # 1. project values.yaml (picked up by helm automatically)
+        # 2. project cluster values.yaml
+        # 3. site.yaml file (on local)
+        # 4. aladdin-config cluster `values.yaml`
+        # 5. aladdin-config cluster namespace `values.yaml`
```
jarojasm95, line 209: 4 and 5 are the new ones

jarojasm95:
@jcwilson I like the idea of supporting yaml all across! I don't think it should be a huge lift, I'll give it a shot

jarojasm95:
Discussed with @jcwilson, for now we'll add this as-is then remove support for a `values` key in `config.json` in a follow up release

jcwilson (approved):

PR #79, opened 2020-11-10
Add aladdin-config default values.yaml and update documentation

jarojasm95:
This PR fixes an inconsistency in #78 where `values.yaml` in aladdin config `default/` directory is not picked up.


jarojasm95:
note: this PR does not remove support for the json-based `values` in aladdin-config so that this can be backwards compatible

aladdin/lib/k8s/helm.py
```diff
@@ -86,16 +86,27 @@ def pull_packages(self, project_name, publish_rules, git_ref, extract_dir):
...
+        Find all possible values yaml files for override in increasing priority
+        Values and overrides are defined/specified in helm args in a specific order
+        1. project values.yaml (picked up by helm automatically)
+        2. project cluster values.yaml
+        3. project cluster namespace values.yaml
+        4. site.yaml file (on local)
+        5. aladdin-config default `values.yaml`
+        6. aladdin-config cluster `values.yaml`
+        7. aladdin-config cluster namespace `values.yaml`
+        8. user passed overrides
```
jarojasm95, line 99: may be worth changing the order of priority here, but that is a larger discussion/effort

jcwilson:
From what I can tell, this won't change any behavior until an aladdin-config directory supplies these values.yaml files (presumably by moving the values out of the `config.json` files. So what really ends up happening is that any values in the `aladdin-config` `config.json` files will take precedence over the `values.yaml` files, since they'll be passed in as the cluster rules `**values` arguments.

When do we deprecate the usage of the `values` key in the `config.json` files?

jcwilson (approved):

jarojasm95:
We can deprecate `values` in `config.json` after migrating `aladdin-config` to use yaml files

PR #83, opened 2021-03-05
[INFRA-563] Update tools for kubernetes upgrade

agianey:
- kubernetes/kubectl -> 1.19.7
- helm -> 3.5.2
- Kops -> 1.19.1
  - required a --admin flag to be passed on exporting the kube config
- Docker -> 20.10.2
- kubernets python lib -> 12.0.01
  - needed some api versions updated
- minikube -> 1.17.1


scripts/infra_k8s_check.sh
```diff
@@ -14,13 +14,13 @@ set -eu -o pipefail
...
 #- from : https://github.com/docker/docker-ce/releases
-VERSION_DOCKER="18.09.7"
+VERSION_DOCKER="20.10.2"
 #- from : https://github.com/kubernetes/kubernetes/releases
-VERSION_KUBECTL="1.15.6"
+VERSION_KUBECTL="1.19.7"
 #- from : https://github.com/kubernetes/helm/releases
-VERSION_HELM="3.3.4"
+VERSION_HELM="3.5.2"
 FULL_VERSION_VIRTUALBOX="6.0.14r133895"
```
jarojasm95, line 24: maybe scope creepy, but this one could go to `6.1.18`, feel free to leave as-is if you have any concerns

jarojasm95 (approved):

Dockerfile
```diff
@@ -38,29 +38,30 @@ RUN apt-get update && \
 
 RUN pip install --no-cache-dir pip==20.2.3
 
+# Update all needed tool versions here
+
 ARG AWS_IAM_AUTHENTICATOR_VERSION=1.17.9
```
jarojasm95, line 43: this can likely be updated as well, but we can also skip it if we don't want to do it in this PR

PR #84, opened 2021-03-15
Update version in pyproject.toml

jarojasm95:


agianey (approved):

aladdin/__init__.py
```diff
@@ -1,2 +1,3 @@
-VERSION = (1, 19, 7, 0)
+VERSION = (1, 19, 7, 1)
```
jarojasm95, line 1: I guess I could leave it at `.0` too? since there's no code change, but not sure if that would ok with git tags and everything else, so I opted for revving to `.1`

pyproject.toml
```diff
@@ -1,6 +1,7 @@
 [tool.poetry]
 name = "aladdin"
-version = "1.15.6.16"
```
jarojasm95, line 3: this one didn't get bumped, not sure if there's an easy way to keep the two in sync, I added a code comment for now

pyproject.toml
```diff
@@ -1,6 +1,7 @@
...
+# NOTE: Also update version in aladdin/__ini__.py
```
agianey, line 3: nit: should be __init__.py

jarojasm95: thanks!

PR #85, opened 2021-03-17
Support specifying additional values files

jarojasm95:
This PR adds a new option to helm commands `--values-file` by allowing filenames to be passed in, which will be then passed to helm as `--values={file_name}` arguments.


aladdin/commands/publish.py
```diff
@@ -84,8 +84,9 @@ def publish(build_only, build_publish_ecr_only, publish_helm_only):
     d = DockerCommands()
     h = Helm()
 
+    git_hash = Git.get_hash()
     # tags is a list in case we want to add other tags in the future
-    tags = [Git.get_hash()]
+    tags = [git_hash]
```
jarojasm95, line 89: we can likely get rid of the array here, since it's unlikely this will change soon (I think)

aladdin/lib/arg_tools.py
```diff
@@ -55,3 +56,37 @@ def add_command(parser, cmd=bash_cmd, help_msg=bash_help["help"]):
...
+    ),
+)
+
+COMMON_OPTION_PARSER = argparse.ArgumentParser(add_help=False)
+COMMON_OPTION_PARSER.add_argument(
+    "--namespace",
+    "-n",
+    default=os.getenv("NAMESPACE", "default"),
+    help="namespace name, defaults current: [{}]".format(os.getenv("NAMESPACE", "default")),
+)
```
jarojasm95, line 92: abstracted common arguments into these base parsers. I opted out of updating all the places where we could use `COMMON_OPTION_PARSER` for now (since it could be used in _all_ the python commands).

aladdin/lib/k8s/helm.py
```diff
@@ -34,6 +35,8 @@ def publish(self, project_name, publish_rules, chart_path, hash):
             "package",
             "--version",
             version,
+            "--app-version",
+            git_hash,
```
jarojasm95, line 39: seems useful, not needed, just thought it was a nice addition

aladdin/lib/k8s/helm.py
```diff
@@ -69,6 +72,10 @@ def pull_packages(self, project_name, publish_rules, git_ref, extract_dir):
             if obj.key.endswith(f".{git_ref}.tgz")
         ]
 
+        if not package_keys:
+            logger.error(f"No published charts found for: {project_name} at {git_ref}")
+            sys.exit(1)
```
jarojasm95, line 77: early check to catch the case where `publish` has not happened yet, otherwise you get a verbose helm error instead

aladdin/commands/deploy.py
```diff
@@ -120,8 +100,16 @@ def deploy(
...
+        values_files = [
+            f"--values={file_name}" for file_name in set_override_values
+            if "=" not in file_name
+        ]
         # Update with --set-override-values
-        value_overrides = {k: v for k, v in (value.split("=") for value in set_override_values)}
+        value_overrides = {
+            k: v for k, v in
+            (value.split("=") for value in set_override_values if "=" in value)
+        }
```
jcwilson, line 112: i'm always a fan of comprehensions, but maybe this is better expressed as

```
value_overrides = dict(value.split("=") for value in set_override_values if "=" in value)
```

jcwilson, line 112: I'm not sure I'm a big fan of mixing argument types like this. I can't imagine there's a realistic danger of a values file having an `=` character in it, but that's a risk with this current implementation.

I think the larger issue is that it seems fairly awkward to convey and utilize. Could we add a new argument like `--values-files`?

aladdin/lib/arg_tools.py
```diff
@@ -55,3 +56,37 @@ def add_command(parser, cmd=bash_cmd, help_msg=bash_help["help"]):
             sub_parser.set_defaults(func=lambda args: bash_wrapper())
         commands.append((bash_cmd, add_command))
     return commands
+
+
+HELM_OPTION_PARSER = argparse.ArgumentParser(add_help=False)
+HELM_OPTION_PARSER.add_argument(
+    "--chart", help="which chart in the project to use, defaults to the project name"
+)
```
jcwilson, line 64: too bad we have to lose the more specified multi-chart functionality for the `start` command. The docs need to be updated to reflect this change.

aladdin/commands/publish.py
agianey: agreed, I think this was here because way back in the day we used to tag with date haha (which we quickly removed)

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "aladdin"
-version = "1.19.7.3"
+version = "1.19.7.4"
```
agianey, line 3: wow! we're already on x.4?? it's only been a week ;)

agianey: We need to update the aladdin/__init__.py file too right?

agianey (commented):
LGTM when we update the aladdin/__init__.py file and after josh's recommendations are added

pyproject.toml
jarojasm95: haha 😉 I updated the `init.py` file so it does not need to be updated anymore! it uses `importlib` to figure out the package version instead 🙂

jarojasm95:
thanks for looking guys 🙂 I'll address the comments soon.

aladdin/commands/publish.py
```diff
@@ -84,27 +84,24 @@ def publish(build_only, build_publish_ecr_only, publish_helm_only):
     d = DockerCommands()
     h = Helm()
 
-    # tags is a list in case we want to add other tags in the future
-    tags = [Git.get_hash()]
```
jarojasm95, line 88: got rid of this list

aladdin/lib/project_conf.py
```diff
@@ -59,7 +59,7 @@ def name(self):
 
     @property
     def mount_path(self):
-        # Under aladdin, we are using /minikube_root , but under minikube it's /
+        # Under aladdin, we are using /aladdin_root , but under minikube it's /
```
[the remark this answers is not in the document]
jcwilson: I have some local changes for this that I can push up after this PR that help with minikube-less aladdin (on windows, at least)

aladdin/lib/arg_tools.py
jcwilson: I've actually been making use of it lately. Would it be difficult to keep?

jarojasm95: not at all, turned out to be pretty simple actually, done!

aladdin/commands/deploy.py
jarojasm95: done!

aladdin/lib/arg_tools.py
```diff
@@ -25,14 +25,15 @@ def get_current_namespace():
...
 
 def add_namespace_argument(arg_parser):
-    namespace_def = get_current_namespace()
     arg_parser.add_argument(
         "--namespace",
         "-n",
-        default=namespace_def,
-        help="namespace name, defaults to default current : [{}]".format(namespace_def),
+        default=CURRENT_NAMESPACE,
+        help=f"namespace name, defaults to default current : [{CURRENT_NAMESPACE}]",
```
abandali, line 36:
```suggestion
        help=f"namespace name, defaults to: [{default}]",
```

abandali (approved):
looks great, 1 minor (nit) suggestion

PR #87, opened 2021-03-30
Add version checks and enforcement

jarojasm95:
This PR adds a some checks to let the user know when there is a newer version of aladdin, aladdin-config, or aladdin-plugins. All these checks use git to compare hashes of the repo's tags (the tag is already configured in aladdin-config). This PR also adds a new config option "enforce_version" (defaulting to false) which requires that the user is up-to-date in order to execute aladdin commands.

### DIFF
--- a/aladdin/lib/version_check.py
+++ b/aladdin/lib/version_check.py
@@ -1,0 +1,50 @@
+import logging
+import os
+import sys
+
+from aladdin import __version__
+from aladdin.lib import in_aladdin_container, utils
+from aladdin.lib.git import Git
+from aladdin.config import load_config, PROJECT_ROOT
+
+
+def check_latest_version():
+    if in_aladdin_container():
+        return
+    config = load_config()
+    repo = config["aladdin"]["repo"]
+    tag = config["aladdin"]["tag"]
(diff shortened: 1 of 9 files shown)


aladdin/lib/version_check.py
```diff
@@ -0,0 +1,63 @@
...
+    git_url = f"git@github.com:{repo}.git"
+
+    _aladdin_version_check(tag, git_url, enforce_version=enforce_version)
+    _config_version_check(tag, enforce_version=enforce_version)
+    _plugins_version_check(tag, enforce_version=enforce_version)
+
+
+def _aladdin_version_check(tag, git_url, enforce_version=False):
+    live = Git.get_hash_ls_remote(tag, git_url, "--tags")
+    expected = Git.get_hash_ls_remote(__version__, git_url, "--tags")
```
jarojasm95, line 30: this one is slightly more lax than the other checks since there's no good way to know the current aladdin git hash, so we use the version number and check that git hash

abandali (commented):
this looks promising/beneficial - not reviewing since it's stale, but happy to re-reviews once it's refreshed (since mount is gone maybe remove `DEV` altogether, rebase etc.)

PR #88, opened 2021-06-10
Linux NFS mounts: tweaks for compatibility and improved security

daniel-rollings-fivestars:
This PR starts with a small patch to detect and use, in descending order of preference:

/srv/nfs/home
/srv/home
/home

...and mount to /home on the container.

- Daniel Rollings

### DIFF
--- a/scripts/install_nfs_mounts.sh
+++ b/scripts/install_nfs_mounts.sh
@@ -6,7 +6,16 @@ case "$OSTYPE" in
         mount_dir="/c/Users"
         ;;
     linux*)
-        export_dir="/home"
+        # We should prefer more hardened Linux conventions if we're to use NFS.
+        # In fact, we're better off not using NFS with its security hazards!
+        # - Daniel Rollings
+        if [ -d "/srv/nfs/home" ]; then
+            export_dir="/srv/nfs/home"
+        elif [ -d "/srv/home" ]; then
+            export_dir="/srv/home"
+        else
+            export_dir="/home"
+        fi
         mount_dir="/home"
         ;;
     *)
--- a/scripts/install_nfs_mounts.sh
+++ b/scripts/install_nfs_mounts.sh
@@ -14,6 +14,7 @@ case "$OSTYPE" in
             export_dir="/home"
         fi
         mount_dir="/home"
+        echo "### ALERT: $0 using export_dir ${export_dir}, mount_dir ${mount_dir}"
         ;;
     *)
         export_dir="/Users"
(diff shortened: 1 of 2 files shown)


daniel-rollings-fivestars:
This fix can expedite my onboarding and I would like to review the architecture of Aladdin and KDEV further before making further changes.  

Squash merge recommended!

scripts/install_nfs_mounts.sh
```diff
@@ -6,7 +6,13 @@ case "$OSTYPE" in
...
         ;;
     linux*)
-        export_dir="/home"
+        if [ -d "/srv/nfs/home" ]; then
+            export_dir="/srv/nfs/home"
+        elif [ -d "/srv/home" ]; then
+            export_dir="/srv/home"
+        else
+            export_dir="/home"
+        fi
```
jarojasm95, line 15: @ptramsey I think your input would be valuable here, this seems fine to me, but I am not too familiar with linux dev workflows or what the "standard" is.

@daniel-rollings-fivestars Do you think maybe a small "echo: 'your nfs home export is: <dir>'` would be valuable here? I can see this possibly being confusing if aladdin picks a different dir other than "/home" (even if the others are more secure)

daniel-rollings-fivestars: Agreed, and though I'm not sure it's enough notification, here's a start.

jcwilson:
>Squash merge recommended!

and required! :)

scripts/install_nfs_mounts.sh
daniel-rollings-fivestars: I've found the most significant blockers driving this PR.  Some are alluded to here:

https://confluence.fivestars.com/display/ENG/Running+Aladdin+on+Linux

But most significantly, hard-coded 192.168.99.1 addresses in Aladdin do not match what got set up in Docker.  This needs to be corrected for anyone who's not running in VirtualBox.  Perhaps this script needs to source configuration variables from a file set to current defaults, but the --init can provision correctly detected ones for this use case?

daniel-rollings-fivestars:
An additional fix for cases where aladdin seems to attempt an NFS mount despite NFS=0, and the script stops there.  At least with || true, we get a failure on a spurious operation but the rest of the script executes as necessary.

daniel-rollings-fivestars:
Unfortunately, as of this writing, Aladdin will still attempt to rely on NFS with minikube even if configured not to do so, and while it will initialize a cluster properly, running "aladdin build" in the pay repo yields:

`Error: Unable to find environment file /root/aladdin-config/minikube/env.sh for specified cluster minikube`

If minikube is to be deprecated soon in favor of k3d, I am prepared to move my bandwidth into k3d migration, and automating it for others.

abandali:
closing since author left, can always re-open if deemed beneficial

PR #89, opened 2021-06-21
Add k3d support and deprecate minikube

agianey:
- Deprecate minikube for k3d
- remove --with-mount functionality (we will be recommending okteto instead)
- Update documentation


aladdin-container.sh
```diff
@@ -100,17 +100,15 @@ function environment_init() {
 
         if "$IS_LOCAL"; then
             mkdir -p $HOME/.kube/
-            cp $HOME/.kube_local/config $HOME/.kube/config
-            # Replace e.g. /Users/ptr/.minikube with /root/.minikube in the minikube conf dir path
-            # (since that's where it will be mounted within the aladdin container)
-            sed 's|: \?.*\.minikube|: /root/.minikube| ; /: \/root\/.minikube/ s|\\|/|g' $HOME/.kube_local/config > $HOME/.kube/config
+            sed "s;https://0.0.0.0:$K3D_API_PORT;https://host.docker.internal:$K3D_API_PORT;g" $HOME/.kube_local/config > $HOME/.kube/config
```
agianey, line 103: this is the address of the apiserver from within the aladdin docker container (host.docker.internal)

aladdin.sh
```diff
@@ -88,6 +65,25 @@ function check_cluster_alias() {
     fi
 }
 
+function get_k3d_ports() {
```
agianey, line 68: Configurable ports for k3d startup. Did not want to default to using port 80 since that may be in use on the host already, but this allows people the option of overriding to that, which allows them to access services locally using port 80

aladdin.sh
```diff
@@ -121,112 +115,25 @@ function check_and_handle_init() {
...
-
-            if test -n "$value"; then
-                minikube config set "$minikube_key" "$value" > /dev/null
-            fi
-        done
-    fi
+function _start_k3d() {
+    # start k3d cluster
+    k3d cluster create LOCAL --api-port "$K3D_API_PORT" -p "$K3D_SERVICE_PORT:80@loadbalancer" \
+        --k3s-server-arg "--tls-san=host.docker.internal" --image "rancher/k3s:v$KUBERNETES_VERSION-k3s1"
```
agianey, line 125: tls-san argument needed to that we don't get certificate issues when hitting the kubernetes api server from within the docker container

aladdin/bash/container/get-dashboard-url/get-dashboard-url
```diff
@@ -1,31 +0,0 @@
-#!/usr/bin/env bash
```
agianey, line 1: the dashboard isn't available by default in k3d. i've moved that functionality to an internal plugin

aladdin/bash/container/host/host
```diff
@@ -10,9 +10,8 @@ function host {
 		# On mac/linux :
 		# - sudo vim /etc/hosts
 		# and add this:
-		$MINIKUBE_IP minikube
 	EOF
-    kubectl get ingress ingress -o=json 2>/dev/null | jq '.spec.rules[]?.host' --raw-output | grep -v '^$' | xargs -I % echo "$MINIKUBE_IP %"
+    kubectl get ingress ingress -o=json 2>/dev/null | jq '.spec.rules[]?.host' --raw-output | grep -v '^$' | xargs -I % echo "127.0.0.1 %"
```
agianey, line 14: local services are available over 127.0.0.1 now

aladdin/commands/build.py
```diff
@@ -18,5 +19,9 @@ def build_args(args):
 
 @container_command
 def build(build_args):
+    tag = "local"
+    k3d = K3d()
     pc = ProjectConf()
-    pc.build_docker(env={"HASH": "local"}, build_args=build_args)
+    pc.build_docker(env={"HASH": tag}, build_args=build_args)
+    images = pc.get_docker_images()
+    k3d.import_images([f"{image}:{tag}" for image in images])
```
agianey, line 27: locally built images aren't available to k3d without importing them. This causes `aladdin build` to take a little bit longer (~15secs/image)

aladdin/commands/restart.py
```diff
@@ -12,19 +12,13 @@ def parse_args(sub_parser):
         dest="charts",
         help="Start only these charts (may be specified multiple times)",
     )
-    subparser.add_argument(
```
agianey, line 15: deprecating --with-mount functionality. will recommend people use okteto instead

aladdin/commands/build.py
jcwilson: 15 seconds, ouch

aladdin/commands/restart.py
jcwilson: Will that documentation be included in this PR at some point?

jcwilson:
I'll try to find some time to try this out on windows/wsl2 (basically linux). Anything I should watch out for?

aladdin/commands/restart.py
agianey: I'm planning on adding plugins for okteto support with some internal documentation

agianey:
I'll have some thorough guinea pig steps later this week. If you're super eager though, i think you'll just need to check out to the INFRA-743 branches in aladdin, aladdin-config, and aladdin-plugins, and update your aladdin tag to use `k3d`

aladdin/commands/build.py
agianey: the plus side is though that we should be building way less, since the images will stay on the host docker instance

aladdin/commands/restart.py
jarojasm95: do you think there's no longer value in supporting mounts given okteto? 
and if so then maybe it'd be worth adding the okteto plugins into core aladdin

aladdin/commands/start.py
```diff
@@ -79,9 +71,7 @@ def start(
     # Start command values
     values = {
         "deploy.imageTag": "local",
-        "deploy.mountPath": pc.mount_path,
         "deploy.namespace": namespace,
-        "deploy.withMount": with_mount,
```
jarojasm95, line 84: I guess this shouldn't break too many helm charts (if any) since a missing value will just be treated as "falsy"

jarojasm95 (commented):
great to see this! left some small comments

jarojasm95:
extra note: does it make sense to get rid of the `aws` secret created in `_replace_aws_secret()`? I think that was only to enable image pulling from ECR - but sounds like that now requires a `docker pull` and a `k3d image import`

aladdin/commands/restart.py
agianey: Yeah I wasn't sure which way to go on this. I ended up deciding on aladdin not being okteto aware (i think we felt similarly about vault), and the okteto plugin is currently fivestars specific

jarojasm95 (commented):
One thing I'm noticing is that aladdin no longer ships with a docker engine, before we would ship it along with minikube.

I think this is a *good* change, since we'll just let the user handle that portion - but I think we should document this somewhere AND we can also remove the installation of the docker client in the infra script

jarojasm95:
An after thought to that^ is that since docker will be available on the host - a lot of aladdin functionality can operate on the host too, which might make aladdin more performant down the line! 🥳

jarojasm95 (approved):

PR #91, opened 2021-07-12
[ALDN-220] Enable ssh agent access in aladdin container

jcwilson:
Mounting the host's `.ssh` directory into the aladdin container at `/root/.ssh` can be problematic
due to ssh's strict requirements regarding file permissions for that directory.

Docker volume mounts set their uid and gid to the host user's uid/gid values. This makes it
difficult for the `root` user to have access to these files without using setuid/setgid tricks, and
in some cases even that is insufficient.

These changes allow a host running an ssh agent with active identities to give the aladdin container
access to the agent by mounting the `SSH_AUTH_SOCK` file. The container's ssh client can then
communicate over the mounted socket, thus eliminating the need to mount the `.ssh` directory into
the container at all.

The behavior will be to default to the `.ssh` directory mount, but utilize the ssh agent socket if the
aladdin config `ssh.agent` is set to `true`.

[ALDN-220](https://jira.fivestars.com/browse/ALDN-220)

### DIFF
--- a/.githooks/lib/trufflehog-excluded.txt
+++ b/.githooks/lib/trufflehog-excluded.txt
@@ -1,1 +1,2 @@
+commands/bash/container/get-dashboard-url/get-dashboard-url
 aladdin/bash/container/get-dashboard-url/get-dashboard-url
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 *.egg-info
 .venv
 dist
+.githooks/.collections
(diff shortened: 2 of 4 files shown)


.githooks/lib/trufflehog-excluded.txt
```diff
@@ -1 +1,2 @@
+commands/bash/container/get-dashboard-url/get-dashboard-url
```
jcwilson, line 1: This file no longer exists in the repo, but trufflehog still complains about it. Any ideas on whether we'll need to keep this line around forever?

agianey: That is strange 🤔

agianey (approved):
LGTM - should we add some documentation around enabling the ssh forwarding to the README?

PR #92, opened 2021-08-09
Add flag to enable waiting for a cert to be available

jarojasm95:
This PR adds a `--wait` flag to the `get-certificate` command

Additionally it adds a util function to get the current namespace

### DIFF
--- a/aladdin/commands/get_certificate.py
+++ b/aladdin/commands/get_certificate.py
@@ -18,20 +18,24 @@ def parse_args(sub_parser):
     )
     subparser.add_argument(
         "--wait",
-        action="store_true",
+        type=int,
+        default=0,
         dest="wait",
-        help="Wait for certificate to be ready",
+        help="Seconds to wait for certificate to be ready (use -1 to wait forever)",
     )
 
 
 @container_command
 @expand_namespace
-def get_certificate(namespace: str, for_cluster: bool = False, wait: bool = False):
+def get_certificate(namespace: str, for_cluster: bool = False, wait: int = 0):
     cr = cluster_rules(namespace=namespace)
     cert = None
+    timeout_start = time.time()
     while not cert:
         cert = cr.cluster_certificate_arn if for_cluster else cr.service_certificate_arn
         if not wait:
+            return cert
+        if wait > 0 and time.time() - timeout_start > wait:
             return cert
         if not cert:
             time.sleep(10)
--- a/aladdin/commands/get_certificate.py
+++ b/aladdin/commands/get_certificate.py
@@ -1,12 +1,14 @@
+import time
+
 from aladdin.lib.cluster_rules import cluster_rules
-from aladdin.lib.arg_tools import add_namespace_argument, container_command
+from aladdin.lib.arg_tools import add_namespace_argument, container_command, expand_namespace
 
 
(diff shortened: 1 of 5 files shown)


agianey (commented):
looks good - one q:
how do we want this to behave if for example ACM is down? should we have some kind of max wait? or do we want to hang indefinitely

jarojasm95:
great point @agianey!

I've updated the PR so we don't wait indefinitely - not sure if we want to make the wait time configurable (instead of a hard 10 secs) - may not be of much importance for now

agianey (approved):

aladdin/lib/arg_tools.py
```diff
@@ -4,12 +4,46 @@
...
+        namespace = "default"
+    return namespace
+
+
+def expand_namespace(func=None):
+    """
+    Decorator to expand a Namespace obj into keyworded arguments
+    """
+
+    @functools.wraps(func)
```
jcwilson, line 33: does this work if `func` `None`?

aladdin/lib/arg_tools.py
```diff
@@ -4,12 +4,46 @@
...
+    def wrapper(*args, **kwargs):
+        if len(args) == 1 and isinstance(args[0], argparse.Namespace):
+            nmspace = vars(args[0])
+            allowed = list(signature(func).parameters.keys())
```
jcwilson, line 37: super minor: you might be able to get away without the `.keys()`

aladdin/lib/arg_tools.py
```diff
@@ -4,12 +4,46 @@
...
+
+    if not namespace:
+        namespace = "default"
+    return namespace
+
+
+def expand_namespace(func=None):
+    """
+    Decorator to expand a Namespace obj into keyworded arguments
+    """
```
jcwilson, line 31: Maybe let's move this out from in between two other functions that deal with the kubernetes namespace value.

jcwilson:
> I've updated the PR so we don't wait indefinitely - not sure if we want to make the wait time configurable (instead of a hard 10 secs) - may not be of much importance for now

I'd typed my related comment up a while back but only submitted the rest today. I think it's a good idea, but wouldn't block on it.

aladdin/lib/arg_tools.py
jarojasm95: moved :+1:

aladdin/lib/arg_tools.py
jarojasm95: it should - because we have:
```python
if not func:
        return expand_namespace
```

aladdin/commands/get_certificate.py
```diff
@@ -1,26 +1,38 @@
...
+def get_certificate(namespace: str, for_cluster: bool = False, wait: bool = False):
     cr = cluster_rules(namespace=namespace)
-    return cr.cluster_certificate_arn if for_cluster else cr.service_certificate_arn
+    cert = None
+    while not cert:
+        cert = cr.cluster_certificate_arn if for_cluster else cr.service_certificate_arn
+        if not wait:
+            return cert
+        if not cert:
+            time.sleep(10)
```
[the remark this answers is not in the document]
jarojasm95: added extra optional argument

PR #97, opened 2021-08-30
Add compatibility fix for K3D on newer linux kernels

jarojasm95:
More info here: https://k3d.io/faq/faq/#solved-nodes-fail-to-start-or-get-stuck-in-notready-state-with-log-nf_conntrack_max-permission-denied

and here: https://k3d.io/faq/faq/#solved-nodes-fail-to-start-or-get-stuck-in-notready-state-with-log-nf_conntrack_max-permission-denied

### DIFF
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -136,8 +136,16 @@ function check_and_handle_init() {
 
 function _start_k3d() {
     # start k3d cluster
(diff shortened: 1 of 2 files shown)


abandali (approved):
confirmed that this fix is backwards compatible with `v20.10.7` and it works for `v20.10.8` as well

PR #101, opened 2021-09-28
Update Poetry to 1.1.10 (latest)

jarojasm95:
This PR updates [Poetry](https://python-poetry.org/) to the latest version

PR #102, opened 2021-09-28
Add Github Actions Pull Request Workflow

jarojasm95:
This PR adds a workflow for building the docker image on pull requests to master.

In the future we could extend this workflow to do other checks such as: tests, formatting checks, etc

This PR also fixes a build issue missed in #101

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,32 @@
+FROM python:3.8.12-buster as build
+
+WORKDIR /root/aladdin
+
+RUN apt-get update && \
+    apt-get -y --no-install-recommends install \
+    gettext \
+    gcc \
+    g++ \
+    curl
+
+RUN python -m venv /root/.venv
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -12,9 +19,10 @@ RUN pip install --upgrade pip && \
     python install-poetry.py --version $POETRY_VERSION
 ENV POETRY_VIRTUALENVS_CREATE="false"
 ENV POETRY_INSTALLER_PARALLEL="false"
+ENV VIRTUAL_ENV=/root/.venv
```
jarojasm95, line 22: for some reason poetry seems to be needing this in order to find the virtual env

agianey (approved):
looks good! can you add a few comments on what each build step is doign?

abandali (approved):
sweet

PR #105, opened 2021-10-19
Attach label to namespace, to aid in owner identification

abandali:
Extended the `create-namespace` command to slap on a label/tag with the id (from AWS) of the person that created the namespace; this will make it easier when removing potentially orphaned namespaces

### DIFF
--- a/aladdin/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
+++ b/aladdin/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
@@ -46,7 +46,7 @@ function add_aws_assume_role_config {
 
 function usage {
     cat <<-EOF
-		usage: aladdin add-aws-assume-role [-h] role profile_name mfa_enabled duration
+		usage: aladdin add-aws-assume-role-config [-h] role profile_name mfa_enabled duration
 
 		positional arguments:
 		  role                  which role to set aws config for
--- a/aladdin/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
+++ b/aladdin/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
@@ -46,7 +46,7 @@ function add_aws_assume_role_config {
 
 function usage {
     cat <<-EOF
-		usage: aladdin aws-assume-role [-h] role profile_name mfa_enabled duration
+		usage: aladdin add-aws-assume-role [-h] role profile_name mfa_enabled duration
 
 		positional arguments:
 		  role                  which role to set aws config for
--- a/aladdin/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
+++ b/aladdin/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
@@ -59,7 +59,7 @@ function usage {
 	EOF
 }
 
-if [[ $# -eq 0 || "$1" == "-h" || "$1" == "--help" ]]; then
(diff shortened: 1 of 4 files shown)


aladdin/bash/container/add-aws-assume-role-config/add-aws-assume-role-config
```diff
@@ -59,7 +59,7 @@ function usage {
 	EOF
 }
 
-if [[ $# -eq 0 || "$1" == "-h" || "$1" == "--help" ]]; then
+if [[ $# -lt 4 || "$1" == "-h" || "$1" == "--help" ]]; then
```
abandali, line 62: scope creep 😱

aladdin/bash/container/create-namespace/create-namespace
```diff
@@ -19,6 +23,9 @@ function usage {
 
 if [[ $# -eq 0 || "$1" == "-h" || "$1" == "--help" ]]; then
     usage
+elif [[ $# -eq 1 ]]; then
+	owner="$(aws sts get-caller-identity --query 'UserId' --output text | awk -F ':' '{print$2}')"
```
abandali, line 27: I prefer `--query` over `| jq` even though it requires an extra param `--output text`

abandali:
both feedback items incorporated, ran it on KDEV:
```
$ aladdin -c KDEV create-namespace bozo
...
namespace/bozo created
namespace/bozo labeled

KDEV1115:abbas> kubectl describe namespace bozo
Name:         bozo
Labels:       owner=abbas.bandali.ops
Annotations:  <none>
Status:       Active
```

aladdin/bash/container/create-namespace/create-namespace
jarojasm95: what is the value of `owner` if the `aws` command fails? does it make the whole script fail?

aladdin/bash/container/create-namespace/create-namespace
```diff
@@ -3,14 +3,18 @@ set -eu -o pipefail
 
 function create_namespace {
     kubectl create namespace $1
+    kubectl label --overwrite namespace $1 owner=$2
```
jarojasm95, line 6: maybe ignore if there's an error here? (like invalid characters in value, or something of that nature)
```suggestion
    kubectl label --overwrite namespace $1 owner=$2 ||:
```
since it's a non-critical piece of functionality (I think)

jarojasm95 (commented):
mostly looks good!

aladdin/bash/container/create-namespace/create-namespace
abandali: > what is the value of `owner` if the `aws` command fails? does it make the whole script fail?

the script will fail - I think we have this flaw in several places in aladdin - I went ahead and plumbed in a default owner of `nobody` so that the script will not fail and complete gracefuilly:
```
IT2771:aladdin Abbas.Bandali$ aladdin -c KDEV create-namespace bozo
...
/root/aladdin/aladdin/bash/container/create-namespace/create-namespace: line 27: abs: command not found
namespace/bozo created
namespace/bozo labeled

KDEV1115:abbas> kubectl describe namespace bozo
Name:         bozo
Labels:       owner=nobody
Annotations:  <none>
Status:       Active
```

jarojasm95 (approved):
lgtm

PR #106, opened 2021-10-28
Use repository for aladdin config

jarojasm95:
This PR adds a new "user" config option called "config_repo". 

This new config option should be set to the repository where the user has their `aladdin config`. Aladdin will then ensure that this repo is always checked out and up to date before running any command. The config repo is expected to be tagged in a similar manner to aladdin, so that aladdin and config tags always match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ You can add these two commands to your profile file if desired.
 
 You will now need to [create your aladdin configuration](./docs/create_aladdin_configuration.md), and link that to aladdin.
 
-    $ aladdin config set config_dir /path/to/aladdin/configuration
+    $ aladdin config set config_repo git@github.com:{git_account}/{repo}.git
 
 ### Manage configuration
 #### Software dependencies
--- a/README.md
+++ b/README.md
@@ -55,6 +55,10 @@ You will now need to [create your aladdin configuration](./docs/create_aladdin_c
 
     $ aladdin config set config_repo git@github.com:{git_account}/{repo}.git
 
+NOTE:
+
+When doing bootstrapping, initial setup or when developing on the aladdin codebase itself it might be useful to be able to point aladdin to a config on the local development environment (ie not yet in source control). This can be done by setting the `ALADDIN_DEV=true` environment variable and by setting the `config_dir` setting (`aladdin config set config_dir /path/to/your/config`).
+
 ### Manage configuration
 #### Software dependencies
 aladdin installs software it needs like helm, jq, aws cli etc.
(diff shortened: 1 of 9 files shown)


aladdin-container.sh
```diff
@@ -96,13 +96,13 @@ function environment_init() {
...
             kubectl config set-context "$NAMESPACE.$CLUSTER_NAME" --cluster "$CLUSTER_NAME" --namespace="$NAMESPACE" --user "$CLUSTER_NAME"
-            kubectl config use-context "$NAMESPACE.$CLUSTER_NAME"            
+            kubectl config use-context "$NAMESPACE.$CLUSTER_NAME"
         fi
 
         _handle_authentication_config
 
         if $INIT; then
-            $PY_MAIN create-namespace $NAMESPACE
+            $PY_MAIN create-namespace $NAMESPACE || true
```
jarojasm95, line 105: @abandali 
I think we missed the `|| true` in #105 - which means this will fail when the namespace already exists, before it would just ignore that case

aladdin/config.py
```diff
@@ -39,3 +49,104 @@ def load_config_from_file(file):
...
+                    config_repo,
+                    e.stderr.strip() or e.stdout.strip()
+                )
+                return False
+
+    os.environ["ALADDIN_CONFIG_DIR"] = str(remote_config_path)
+
+    if strtobool(os.getenv("ALADDIN_DEV", "false")) and config_dir and os.path.isdir(config_dir):
+        """
+        Allow aladdin developers to use a custom config
```
abandali, line 148: might be helpful to add some notes in the README about this customization via `$ALADDIN_DEV`, now that `--dev` is gone

README.md
```diff
@@ -53,7 +53,7 @@ You can add these two commands to your profile file if desired.
 
 You will now need to [create your aladdin configuration](./docs/create_aladdin_configuration.md), and link that to aladdin.
 
-    $ aladdin config set config_dir /path/to/aladdin/configuration
+    $ aladdin config set config_repo git@github.com:{git_account}/{repo}.git
```
abandali, line 56: a note here for aladdin devs to be able to override using `$ALADDIN_DEV` would be helpful

abandali (commented):
LGTM - a small suggestion

README.md
jarojasm95: done

aladdin/config.py
jarojasm95: done 👍

abandali (approved):
awesome !

PR #109, opened 2021-11-08
Align configs with live version

sergey-korenets-fivestars:


config-example/LOCAL/config.json
```diff
@@ -1,5 +1,6 @@
 {
     "is_local": true,
+    "is_prod": false,
```
abandali, line 3: technically this is not needed as the default is set in `aladdin.sh` if the value extraction fails: https://github.com/fivestars-os/aladdin/blob/1.19.7.14/aladdin.sh#L157-L160

notice that the sample PROD config has the correct setting: https://github.com/fivestars-os/aladdin/blob/1.19.7.14/config-example/CLUSTERPROD/config.json#L6

is there a compelling reason for setting this in this example LOCAL config?

config-example/default/config.json
```diff
@@ -6,5 +6,6 @@
     "cluster_init": [
         {"project": "dashboard", "ref": "v1.0.0", "namespace": "kube-system"}
     ],
-    "dual_dns_prefix_annotation_name": "dns-name"
+    "dual_dns_prefix_annotation_name": "dns-name",
+    "bastion_account_enabled": false
```
abandali, line 10: good catch !

config-example/LOCAL/config.json
sergey-korenets-fivestars: @abandali , thanks, just verified, it works fine without this option.

abandali (approved):
LGTM

PR #110, opened 2021-11-08
Use aladdin image matching current version

jarojasm95:
This PR updates aladdin to use the current package version to pull the corresponding docker image. This ensures that the host code and the container code are always in sync, otherwise there could be breaking changes between the two.

### DIFF
--- a/aladdin/env.py
+++ b/aladdin/env.py
@@ -1,0 +1,116 @@
+import os
+import os
+import pathlib
+import subprocess
+
+from jmespath import search
+
+from aladdin import __version__, config
+from aladdin.lib import logging
+from aladdin.lib.utils import working_directory
+
+logger = logging.getLogger(__name__)
+
+
+def configure_env():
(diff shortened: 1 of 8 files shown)


aladdin/env.py
```diff
@@ -0,0 +1,134 @@
...
+    if not (os.getenv("ALADDIN_IMAGE") and config.ALADDIN_DEV):
+        # Set ALADDIN_IMAGE from the "aladdin.repo" config (or config.ALADDIN_DOCKER_REPO) and
+        # use the aladdin version as the image tag
+        os.environ["ALADDIN_IMAGE"] = "{}:{}".format(
+            search("aladdin.repo", config.load_config()) or config.ALADDIN_DOCKER_REPO,
+            __version__
+        )
+
+
+def set_config_path() -> bool:
```
jarojasm95, line 35: this was mostly just moved from config.py - added some docstrings

PR #111, opened 2021-11-11
Make namespace creation declarative

sergey-korenets-fivestars:
1. Fix the permanent error from aladdin:
`Error from server (AlreadyExists): namespaces "default" already exists`
2. Keep aladdin output dry

### DIFF
--- a/aladdin/bash/container/create-namespace/create-namespace
+++ b/aladdin/bash/container/create-namespace/create-namespace
@@ -2,7 +2,7 @@
 set -eu -o pipefail
 
 function create_namespace {
-    kubectl create namespace $1
+    kubectl create namespace $1 --dry-run=client -o yaml | kubectl apply -f -
     kubectl label --overwrite namespace $1 owner=$2 || true
 }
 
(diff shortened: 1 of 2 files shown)


aladdin/bash/container/create-namespace/create-namespace
```diff
@@ -2,7 +2,7 @@
 set -eu -o pipefail
 
 function create_namespace {
-    kubectl create namespace $1
+    kubectl create namespace $1 --dry-run=client -o yaml | kubectl apply -f -
```
jarojasm95, line 5: I ran this and got:
> Warning: resource namespaces/default is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.


so it seems we're getting rid of one error but now we'll have this permanent warning?

sergey-korenets-fivestars: @jarojasm95 , i don't think so. During next runs, you shouldn't see this warning.

jarojasm95 (approved):

PR #112, opened 2021-11-11
INFRA-982: add service.clusterName

abandali:
This is an alternative to setting it via [aladdin-config](https://github.com/fivestars/aladdin-config/pull/50) - seems cleaner, but open to feedback if anyone has a preference


jarojasm95 (approved):

jarojasm95:
small nit: not sure if we'll want to add the same thing to the `start()` command?

jarojasm95 (approved):

PR #113, opened 2021-11-17
Enable auto-updating as part of infra checks

jarojasm95:
This PR adds automatically updating aladdin to the latest released version, this piggy backs off the `infra_k8s_check` script to run this periodically as part of the initialization logic


scripts/install-aladdin.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+ALADDIN_VERSION=${ALADDIN_VERSION:-$(\
+curl -s https://api.github.com/repos/fivestars-os/aladdin/releases/latest | \
+python3 -c 'import json,sys;print(json.load(sys.stdin)["tag_name"])' || \
```
jarojasm95, line 5: could also use `jq` - but since aladdin installs `jq` it seems reasonable to assume that `jq` might not always be available at install time, however we do expect python to always be available

abandali (approved):
this is awesome

pyproject.toml
```diff
@@ -16,7 +16,6 @@ boto3 = "1.18.28"
 cached-property = "1.5.1"
 jmespath = "0.10.0"
 kubernetes = "12.0.01"
-orderedset = "^2.0.3"
```
jarojasm95, line 19: removed this as it was only being used in one place

PR #115, opened 2021-12-07
Avoid duplicate repo prefix

jarojasm95:
Trying to automatically inject `repo` values into the user aladdin config can result in a value with a duplicated prefix, example: `"plugin_repo": "git@github.com:git@github.com:fivestars/aladdin-plugins.git"`.

This PR adds a workaround to prevent that

### DIFF
--- a/aladdin-container.sh
+++ b/aladdin-container.sh
@@ -102,7 +102,7 @@ function environment_init() {
         _handle_authentication_config
 
         if $INIT; then
-            $PY_MAIN create-namespace $NAMESPACE || true
+            $ALADDIN_DIR/aladdin/bash/container/create-namespace/create-namespace $NAMESPACE || true
             $PY_MAIN namespace-init --force
         fi
     fi
--- a/aladdin/env.py
+++ b/aladdin/env.py
@@ -90,7 +90,7 @@ def set_repo_path(env_key: str, dir_key: str, repo_key: str, required: bool = Fa
             return False
         try:
             remote = subprocess.run(
-                "git remote get-url origin".split(),
+                "git config --get remote.origin.url".split(),
                 check=True,
                 capture_output=True,
                 encoding="utf-8",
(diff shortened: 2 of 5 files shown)


aladdin/env.py
```diff
@@ -100,7 +100,9 @@ def set_repo_path(env_key: str, dir_key: str, repo_key: str, required: bool = Fa
             logger.error(err_message)
             return False
         *_, git_account, repo = remote.split("/")
-        repo_value = f"git@github.com:{git_account}/{repo}"
```
jarojasm95, line 103: I am unsure why this would result in a duplicated `git@github.com:`, so this is more of a workaround instead of a fix

aladdin/lib/k8s/kubernetes.py
```diff
@@ -33,7 +33,7 @@ def __init__(
         self.kubeconfig = kubeconfig or os.getenv("KUBECONFIG")
         try:
             config.load_kube_config(self.kubeconfig)
-        except IOError:
```
jarojasm95, line 36: small bug fix

aladdin-container.sh
```diff
@@ -102,7 +102,7 @@ function environment_init() {
         _handle_authentication_config
 
         if $INIT; then
-            $PY_MAIN create-namespace $NAMESPACE || true
+            $ALADDIN_DIR/aladdin/bash/container/create-namespace/create-namespace $NAMESPACE || true
```
jarojasm95, line 105: `create-namespace` is a bash plugin, but we are already inside the container, so we need to call the script directly (can't use `PY_MAIN`)

scripts/install-aladdin.sh
```diff
@@ -5,6 +5,14 @@ curl -s https://api.github.com/repos/fivestars-os/aladdin/releases/latest | \
 python3 -c 'import json,sys;print(json.load(sys.stdin)["tag_name"])' || \
 echo -n "master")}
 
-python3 -m pip install --user --upgrade pipx
-python3 -m pipx ensurepath
-python3 -m pipx install git+https://github.com/fivestars-os/aladdin.git@${ALADDIN_VERSION} --force
```
jarojasm95, line 10: this had very verbose output, the change makes it so we only run these commands when necessary

aladdin/env.py
abandali: yeah, I just ran into this .. weird how it's happening

abandali (approved):
LGTM

PR #116, opened 2021-12-14
Fix namespace creation label & hosted zone delegation

jarojasm95:
This PR fixes two bugs:
- `create-namespace` plugin throws an error when we are unable to get an owner using awscli
- A hosted zone is not setup correctly when it is created by the `get-certificate` command

### DIFF
--- a/aladdin/bash/container/create-namespace/create-namespace
+++ b/aladdin/bash/container/create-namespace/create-namespace
@@ -3,7 +3,9 @@ set -eu -o pipefail
 
 function create_namespace {
     kubectl create namespace $1
-    kubectl label --overwrite namespace $1 owner=$2 || true
+    if [ -z "$2" ]; then
+    	kubectl label --overwrite namespace $1 owner=$2 || true
+    fi
 }
 
 function usage {
@@ -15,7 +17,7 @@ function usage {
 		  owner                 the owner of the namespace (applied as a label)
 		                        ideally using the format First.Last
 		                        if not supplied then retrieved from `whoami`
-          
+
 		optional arguments:
 		  -h, --help            show this help message and exit
 	EOF
@@ -24,7 +26,7 @@ function usage {
 if [[ $# -eq 0 || "$1" == "-h" || "$1" == "--help" ]]; then
     usage
 elif [[ $# -eq 1 ]]; then
-	owner="$(aws sts get-caller-identity --query 'UserId' --output text | awk -F ':' '{print$2}' || echo nobody)"
+	owner="$(aws sts get-caller-identity --query 'UserId' --output text | awk -F ':' '{print$2}' || echo '')"
     create_namespace $1 $owner
 else
     create_namespace $@
--- a/aladdin/lib/aws/certificate.py
+++ b/aladdin/lib/aws/certificate.py
@@ -5,7 +5,7 @@ import logging
 import re
 from hashlib import md5
 
-from aladdin.lib.aws.dns_mapping import create_hostedzone, fill_dns_dict, get_hostedzone
+from aladdin.lib.aws.dns_mapping import fill_hostedzone
 
 
 def search_certificate_arn(boto_session, dns_name):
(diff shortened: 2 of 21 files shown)


aladdin/bash/container/create-namespace/create-namespace
```diff
@@ -24,7 +26,7 @@ function usage {
 if [[ $# -eq 0 || "$1" == "-h" || "$1" == "--help" ]]; then
     usage
 elif [[ $# -eq 1 ]]; then
-	owner="$(aws sts get-caller-identity --query 'UserId' --output text | awk -F ':' '{print$2}' || echo nobody)"
+	owner="$(aws sts get-caller-identity --query 'UserId' --output text | awk -F ':' '{print$2}' || echo '')"
```
jarojasm95, line 29: leaving it empty instead of "nobody"

aladdin/bash/container/create-namespace/create-namespace
```diff
@@ -3,7 +3,9 @@ set -eu -o pipefail
 
 function create_namespace {
     kubectl create namespace $1
-    kubectl label --overwrite namespace $1 owner=$2 || true
+    if [ -z "$2" ]; then
+    	kubectl label --overwrite namespace $1 owner=$2 || true
```
jarojasm95, line 7: only set the label if we have an owner (arg 2)

aladdin/lib/aws/dns_mapping.py
```diff
@@ -2,11 +2,14 @@
...
+from aladdin.lib.cluster_rules import ClusterRules
+
 
 def fill_hostedzone(
-    boto_session, cluster_domain_name, namespace_domain_name, hostnames_to_loadbalancers
+    boto_session, hostnames_to_loadbalancers, cluster_domain_name=None, namespace_domain_name=None
 ):
-    # Main DNS is on prod, sub DNS should be on sandbox
+    cluster_domain_name = cluster_domain_name or ClusterRules().cluster_domain_name
+    namespace_domain_name = namespace_domain_name or ClusterRules().namespace_domain_name
```
jarojasm95, line 12: the singleton stuff above is mainly to support this right here, in case the caller does not provide these domain names

aladdin/lib/aws/certificate.py
```diff
@@ -71,38 +71,32 @@ def new_certificate_arn(boto_session, dns_name):
...
             cname_record = {cname_info["Name"][:-1]: cname_info["Value"]}
         except (KeyError, IndexError):
             retry_count += 1
-        else:
-            break
-    if retry_count == max_retries:
+
+    if cname_record:
+        # Create the hosted zone and add the cname used for dns validation
+        fill_hostedzone(boto_session, cname_record)
```
jarojasm95, line 92: this is the core change, previously we used `fill_dns_dict` which correctly setup the CNAME for DNS validation, but it did not correctly setup delegation of the newly created hosted zone. `fill_hostedzone` does all of this

aladdin/lib/aws/dns_mapping.py
```diff
@@ -2,11 +2,14 @@
...
 from datetime import datetime
 
+from aladdin.lib.cluster_rules import ClusterRules
+
 
 def fill_hostedzone(
-    boto_session, cluster_domain_name, namespace_domain_name, hostnames_to_loadbalancers
+    boto_session, hostnames_to_loadbalancers, cluster_domain_name=None, namespace_domain_name=None
 ):
-    # Main DNS is on prod, sub DNS should be on sandbox
```
abandali, line 9: was this comment misplaced (does it need to go somewhere else)?

aladdin/bash/container/create-namespace/create-namespace
abandali: good catch

abandali (commented):
LGTM - do we need to bump the version?

aladdin/lib/aws/dns_mapping.py
jarojasm95: I think this is very specific to Fivestars and is not totally relevant for aladdin

PR #117, opened 2021-12-17
[INFRA-1011] Allow for customizable bash profiles

agianey:
Allows for us to put all handy functions/variables/aliases in a repo and have it available in aladdin bash


jarojasm95 (approved):
nice! should probs bump aladdin version?

jarojasm95 (approved):

PR #118, opened 2022-01-11
Add backport for cached property

jarojasm95:
This PR adds a fix for running aladdin in python < 3.8, `cached_property` was only added to `functools` until python 3.8, so we need a backport in order to support python 3.7


chad-lung-fivestars (commented):
👍

christopher-shubert-sumup (approved):
LGTM

PR #119, opened 2022-01-12
Add check for cgroups v2

jarojasm95:
This PR adds an error message if `docker info` reports that it using cgroups v2. The current image of `k3s` we are using is only compatible with v1

### DIFF
--- a/aladdin.sh
+++ b/aladdin.sh
@@ -114,6 +114,10 @@ function _start_k3d() {
 }
 
 function check_or_start_k3d() {
+    if docker info | grep "Cgroup Version: 2" > /dev/null && [[ "$KUBERNETES_VERSION" == "1.19.7" ]]; then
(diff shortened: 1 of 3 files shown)


aladdin.sh
```diff
@@ -114,6 +114,10 @@ function _start_k3d() {
 }
 
 function check_or_start_k3d() {
+    if docker info | grep "Cgroup Version: 2" > /dev/null && [[ "$KUBERNETES_VERSION" == "1.19.7" ]]; then
+        echo "ERROR: Current version of k3d is not compatible with cgroups v2"
+        echo "ERROR: If using Docker Desktop please downgrade to v4.2.0"
+    fi
```
jarojasm95, line 120: I didn't add an `exit 1` in here in case it's a false positive (not sure if this is an issue in linux/windows as well), mostly just wanted to let the user know it might be an issue

agianey (approved):

PR #122, opened 2022-02-15
Add helm-values command

jarojasm95:
This PR adds a new `helm-values` command that outputs all the helm values to be injected into a specific project and chart. This command is specifically setup to work as a helm downloader plugin.


aladdin/charts/merger/templates/values.yaml
```diff
@@ -0,0 +1 @@
+{{- toYaml .Values -}}
```
jarojasm95, line 1: this is just a dummy helm chart, but it allows us to merge yaml values file in the _exact_ same way that helm does (because we're using helm!)

aladdin/commands/deploy.py
```diff
@@ -86,7 +86,7 @@ def deploy(
             )
             sys.exit(1)
 
-        helm.pull_packages(project, pr, git_ref, tmpdirname)
+        helm.pull_packages(project, pr, git_ref, tmpdirname, chart_name=chart)
```
jarojasm95, line 89: only pull the one we actually want

jarojasm95:
when fully setup as a helm plugin it would look like:
```
helm upgrade --install my-release path/to/my-chart -f aladdin://CLUSTER/my-repo?chart=my-chart&git-ref=main
```

aladdin/commands/helm_values.py
```diff
@@ -0,0 +1,136 @@
...
+                "template",
+                str(PROJECT_ROOT / "aladdin/charts/merger"),
+                *args,
+            ]
+            logging.info("Executing: %s", " ".join(command))
+            subprocess.run(
+                command,
+                capture_output=False,
+                check=True,
+            )
```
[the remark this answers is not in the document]
jarojasm95: good point, I've consolidated the construction of dynamic helm values and the composition of the helm command itself

sergey-korenets-fivestars (approved):

PR #126, opened 2022-03-09
Pin markupsafe to v2.0.1

jarojasm95:
https://github.com/pallets/markupsafe/issues/284

Markupsafe v2.1.0 has a breaking change 😢


nabil-khawaja-fivestars (approved):
Thanks for finding this fix!

PR #127, opened 2022-03-09
Skip clone and checkout repository in helm-values command

jarojasm95:
This PR is a small improvement to the `helm-values` command so that aladdin can skip the "clone" and "checkout" steps if possible

### DIFF
--- a/aladdin/commands/helm_values.py
+++ b/aladdin/commands/helm_values.py
@@ -103,13 +103,14 @@ def clone_and_checkout(repo_name, githash):
         current_repo = Git.get_repo()
         current_hash = Git.get_hash()
 
-    if all([
-        current_hash,
-        current_repo,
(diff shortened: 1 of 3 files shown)


aladdin/commands/helm_values.py
```diff
@@ -106,3 +93,44 @@ def helm_values(
...
+        current_hash = Git.get_hash()
+
+    if (
+        current_hash and
+        current_repo and
+        (not repo_name or current_repo == repo_name) and
+        current_hash == githash and
+        Git.clean_working_tree()
+    ):
+        yield Git.get_base_directory()
```
jarojasm95, line 113: this can be specially useful in CI/CD use-cases, where this will be already executed from the right repo at the right commit

sergey-korenets-fivestars (approved):

PR #129, opened 2022-03-15
Add kubectl aliases

jarojasm95:


Dockerfile
```diff
@@ -83,6 +83,9 @@ RUN curl -L https://istio.io/downloadIstio | ISTIO_VERSION="$ISTIO_VERSION" sh -
 ARG K3D_VERSION=4.4.8
 RUN curl -s https://raw.githubusercontent.com/rancher/k3d/main/install.sh | TAG=v$K3D_VERSION bash
 
+RUN curl -fsSL https://raw.githubusercontent.com/ahmetb/kubectl-aliases/master/.kubectl_aliases -o /etc/profile.d/kubectl_aliases.sh && \
```
jarojasm95, line 86: If there is a better source of aliases that folks know about, I am happy to use whichever

sergey-korenets-fivestars:
Looks funny :) It's like learn new foreign words, especially I like: ksysgdepowidesl.
If we'll use those commands somewhere in automation (not for personal/debugging purpose), I would propose to add link to https://github.com/ahmetb/kubectl-aliases/blob/master/.kubectl_aliases near that command.

sergey-korenets-fivestars (approved):

PR #137, opened 2022-08-15
Upgrade helm, aws-iam-authenticator, awscli

jarojasm95:


Dockerfile
```diff
@@ -80,7 +82,7 @@ ARG KUBE_VERSION=1.19.7
 RUN curl -L -o /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v$KUBE_VERSION/bin/linux/$(dpkg --print-architecture)/kubectl && \
     chmod 755 /usr/local/bin/kubectl
 
-ARG HELM_VERSION=3.8.2
+ARG HELM_VERSION=3.9.2
```
jarojasm95, line 85: upgrading `helm` means we need to upgrade `aws-iam-authenticator` which means we also have to upgrade `awscli` - yikes

jarojasm95:
I've tested this by doing:
```
docker build .
export ALADDIN_IMAGE=<image hash from above>
aladdin -c KDEV -n jose bash
helm ls
# profit!
```

agianey (approved):
Nice!

PR #138, opened 2022-08-19
Add support for multiple values file paths

jarojasm95:
This PR enables a new aladdin config option called `values_files`.

This new option will allow clusters to declare a list of values files that should be picked up from a project helm chart, previously this was inferred from the `cluster_name` config option.

### DIFF
--- a/aladdin/lib/k8s/helm.py
+++ b/aladdin/lib/k8s/helm.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import sys
 from os.path import join
+from typing import List
 
 import yaml
 from botocore.exceptions import ClientError
@@ -89,7 +90,7 @@ class Helm:
                 except FileNotFoundError:
                     pass
 
-    def find_values(self, chart_path, cluster_name, namespace):
+    def find_values(self, chart_path: str, values_files: List[str], namespace: str):
         """
         Find all possible values yaml files for override in increasing priority
         Values and overrides are defined/specified in helm args in a specific order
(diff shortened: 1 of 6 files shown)


aladdin/lib/k8s/helm.py
```diff
@@ -106,20 +107,28 @@ def find_values(self, chart_path, cluster_name, namespace):
         # cluster_name could be an alias, "CLUSTER_CODE" is what is used in aladdin-config
         cluster_code = os.environ["CLUSTER_CODE"]
 
-        cluster_values_path = join(chart_path, "values", f"values.{cluster_name}.yaml")
-        if os.path.isfile(cluster_values_path):
-            logger.info("Found cluster values file")
-            values.append(cluster_values_path)
+        for values_file in values_files:
```
jarojasm95, line 110: the main use case I see for this is multi-region support, so a project could declare:
```
values.CLUSTER.yaml
values.region-1.CLUSTER.yaml
values.region-2.CLUSTER.yaml
```

and aladdin would pick up the right values files based on aladdin-config

aladdin/lib/k8s/helm.py
```diff
@@ -106,20 +107,28 @@ def find_values(self, chart_path, cluster_name, namespace):
...
-            logger.info("Found cluster namespace values file")
-            values.append(cluster_namespace_values_path)
+            cluster_namespace_values_path = join(
+                chart_path, "values", f"values.{values_file}.{namespace}.yaml"
+            )
+            if os.path.isfile(cluster_namespace_values_path):
+                logger.info("Found cluster namespace values file")
+                values.append(cluster_namespace_values_path)
+
+            cluster_custom_values_path = join(
```
jarojasm95, line 123: this also allows for completely arbitrary path structures, example:
```json
"values_files": ["CLUSTER/values.yaml", "CLUSTER/region-2.yaml"]
```
would allow a file structure of:
```
values/
    CLUSTER/
        values.yaml
        region-2.yaml
```

sergey-korenets-fivestars (approved):

PR #145, opened 2022-09-21
Fix `PYTHONPATH` in aladdin auto update

jarojasm95:


scripts/infra_k8s_check.sh
```diff
@@ -461,5 +461,5 @@ function main(){
 
 main
 if ! ${ALADDIN_DEV:-false} ; then
-    "$(cd "$(dirname $0)" ; pwd)"/install-aladdin.sh
+    env -u PYTHONPATH "$(cd "$(dirname $0)" ; pwd)"/install-aladdin.sh
```
jarojasm95, line 464: we want to make sure `PYTHONPATH` is not set for the install script so we don't confuse `pipx`

PR #153, opened 2023-01-26
Bump up kops and kubectl

sergey-korenets-fivestars:
- Bump up kops from 1.23.4 to 1.24.5
 - Bump up kubectl from 1.23.10 to 1.24.10


Dockerfile
```diff
@@ -78,7 +78,7 @@ ARG DOCKER_COMPOSE_2_VERSION=v2.14.2
 RUN curl -L "https://github.com/docker/compose/releases/download/$DOCKER_COMPOSE_2_VERSION/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose-2 && \
     chmod 755 /usr/local/bin/docker-compose-2
 
-ARG KUBE_VERSION=1.23.10
+ARG KUBE_VERSION=1.24.10
```
jarojasm95, line 81: is this needed? we usually try to keep this exactly the same as the kubernetes version we are targeting

jarojasm95 (approved):
seems fine

Dockerfile
jarojasm95: oh I see, also planning to update kubernetes

PR #162, opened 2023-11-13
[INFRA-2155] Removing aladdin dns-sync

vitali-federau-fivestars:
This PR removes aladdin's functionality to create records for R53 on deploy/undeploys/start/end as records are being managed right now by external-dns
I have left the functionality to create records for ACM validation functionality
WIP - not sure if we need to leave the possibility for certificate.py to create Hosted Zones (under fill_hostedzone function) as well, @agianey @jarojasm95 could I ask for your opinion about that please, what do you think, should we leave the creation of HZs for certificates, or we should only allow to create ACM validation records and to remove the ability to create Zones?


agianey:
I think for the HZ stuff, we can get rid of the creation logic, but then the ACM validation should know to use the correct hz

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "aladdin"
-version = "1.25.15.0"
+version = "1.25.16.0"
```
agianey, line 3: we'll want to leave this as 1.25.15 (should match k8s version), and update only the last digit

agianey (commented):
i think sync_dns removal code looks fine. will hold off on approving until we update the cert logic stuff

agianey:
perhaps we should only do the cert validation stuff for dynamic namespaces on KDEV? and then every other cert should be terraformed (like their associated hz is)

vitali-federau-fivestars:
> perhaps we should only do the cert validation stuff for dynamic namespaces on KDEV? and then every other cert should be terraformed (like their associated hz is)

@agianey good catch, but I think it can be an issue to leave dynamic namespaces (and thus Hosted Zones and Certificates) on KDEV, as it will be different from other clusters and will ruin our consistency a little
I was thinking a lot about it and discussed with @sergey-korenets-fivestars about possible ways to handle it, so now I can suggest next:
1. To totally remove the sync-dns function (in this PR already done) so we can disable the possibility of aladdin to create new records
2. To remove the function create_hostedzone to disable the ability to create new hosted zones by aladdin (already done in this PR)
3. To remove the possibility to create and validate ACMs for all clusters and environments (not done in this PR yet as can be discussed)

So in final solution, we will have to create Hosted Zones and ACMs by terraform only on all levels bringing terraform as the only source of trust. For all the records in kdev.awesomestartup.com I will need to update our services to create records in a changed format from loyalty.vitali.kdev.awesomestartup.com to loyalty-vitali.kdev.awesomestartup.com (will need to create a condition in services I assume, this is needed to proceed having HTTPS for records as loyalty.vitali.kdev.awesomestartup.com will not be covered by *.kdev.awesomestartup.com Certificate) and so on KDEV all of the dynamic namespaces/records will be created directly in kdev.awesomestartup.com HZ
The only issue here will be the process of adding new namespaces - it will either create a record in kdev.awesomestartup.com HZ OR we will need to bring the HZ manually with terraform

@agianey @jarojasm95 @sergey-korenets-fivestars what do you think about the above?

agianey (approved):
LGTM! so we'll address the hz stuff in a separate PR, correct?

PR #167, opened 2024-04-29
Skip creation of new hosted zone for ACM certificate requests

jarojasm95:
We no longer use dedicated hosted zones for namespace management - however the certificate command still creates a dedicated hosted zone.

This PR removes the dedicated hosted zone logic


drytdubs (approved):
Passes my high level check; but, I'm not an expert in the consequences of the external_dns changes.

PR #168, opened 2024-05-15
Fix infra-k8s install script versions & architecture

jarojasm95:


nabil-khawaja-fivestars (approved):
Ah thanks for putting this up, was gonna go ahead and do this after we made sure we got everything working properly. There was also something to change in the install-aladdin.sh script to update it to look at main instead of master on line 6.

Also not sure if we need to do any further check for the Mac case if there's an Intel chip vs Apple chip when we do the installs using
> `typeset url="https://dl.k8s.io/v${VERSION_KUBECTL}/kubernetes-client-darwin-amd64.tar.gz"`

vs

> `typeset url="https://dl.k8s.io/v${VERSION_KUBECTL}/kubernetes-client-darwin-arm64.tar.gz"`

and similar case for the helm install, but I think updating these versions for now may help with the aladdin --init defaulting back to the old values for now.

PR #182, opened 2024-10-28
fix: Cast certificate cache path to string

jarojasm95:


aladdin/lib/cache.py
```diff
@@ -38,7 +39,7 @@ def certificate_cache(func):
     @functools.wraps(func)
     @clear_on_error()
     def wrapper(certificate_scope):
-        cache = shelve.open(cache_path)
+        cache = shelve.open(str(cache_path))
```
jarojasm95, line 42: seems like some python versions expect this explicitly as a string instead of a path

aladdin/lib/cache.py
```diff
@@ -20,6 +20,7 @@ def clear_on_error():
     except Exception:
         logging.info("Clearing cache due to error")
         shutil.rmtree(cache_root, ignore_errors=True)
+        raise
```
jarojasm95, line 23: should not swallow errors

nabil-khawaja-fivestars (approved):

PR #185, opened 2024-11-05
fix: Set current working directory in deploy command

jarojasm95:


aladdin/commands/deploy.py
```diff
@@ -70,9 +70,8 @@ def deploy(
         )
         sys.exit(1)
 
-    with clone_and_checkout(git_ref, repo) as tmpdirname:
-        with working_directory(tmpdirname):
-            helm_chart_path = ProjectConf().get_helm_chart_path(chart)
+    with clone_and_checkout(git_ref, repo, cwd=True):
```
jarojasm95, line 73: the helm-values command relies on cwd

PR #196, opened 2025-03-24
fix: Update release

sergey-korenets-fivestars:


.github/workflows/release.yaml
```diff
@@ -47,22 +47,15 @@ jobs:
             type=ref,event=branch
             type=ref,event=tag
 
-      - uses: actions/github-script@v7
```
sergey-korenets-fivestars, line 50: This step fails with error:
```
Error: Unhandled error: SyntaxError: Invalid or unexpected token
```

vitali-federau-fivestars (approved):

PR #201, opened 2025-04-02
fix: Update cluster rules "is local" check

jarojasm95:


.github/workflows/release.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
         with:
           result-encoding: string
           script: |
-            const tags = '${{ steps.meta.outputs.tags }}'.split('\n')
+            const tags = `${{ steps.meta.outputs.tags }}`.split('\n')
```
jarojasm95, line 55: this is a multi-line string which single quotes can't handle well

